PR #9, opened 2021-07-22
sort root directories

ChamseddineBhd:
Signed-off-by: chamseddine <chamseddine.benhamed@ensi-uma.tn>

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -218,7 +218,6 @@ const App = () => {
             sortedData.sort(function (a, b) {
                 return a.elementName.localeCompare(b.elementName);
             });
-            console.log(sortedData);
             setRootDirectories(sortedData);
         });
     };
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -212,11 +212,20 @@ const App = () => {
         connectNotificationsUpdateConfig,
     ]);
 
+    const updateRootDirectories = () => {
+        fetchRootFolders().then((data) => {
+            let sortedData = [...data];
+            sortedData.sort(function (a, b) {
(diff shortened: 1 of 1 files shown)


src/components/app.js
```diff
@@ -212,11 +212,20 @@ const App = () => {
         connectNotificationsUpdateConfig,
     ]);
 
+    const updateRootDirectories = () => {
+        fetchRootFolders().then((data) => {
+            let sortedData = [...data];
+            sortedData.sort(function (a, b) {
+                return a.elementName.localeCompare(b.elementName);
+            });
+            console.log(sortedData);
```
sBouzols: Do we need to keep this log here ?

src/components/app.js
```diff
@@ -212,11 +212,20 @@ const App = () => {
         connectNotificationsUpdateConfig,
     ]);
 
+    const updateRootDirectories = () => {
+        fetchRootFolders().then((data) => {
+            let sortedData = [...data];
+            sortedData.sort(function (a, b) {
```
sBouzols, line 218: Why do we need a sort ? Is directory Tree component not a better place to sort those root directories ?

sBouzols (commented):
not sure if app.js is the best place to add this sort

src/components/app.js
ChamseddineBhd: we need to sort root directories to keep trees in the same order once we delete or insert a new root directory

sBouzols: ok

sBouzols (approved):
ok

PR #10, opened 2021-07-26
add checkboxes to each rows

JacquesBorsenberger:


sBouzols:
Shouldn't we add this functionality in the common-ui/virtualizedTable ? it could be useful elsewhere.

sBouzols (approved):
Ok for me

package.json
```diff
@@ -4,7 +4,7 @@
     "homepage": ".",
     "private": true,
     "dependencies": {
-        "@gridsuite/commons-ui": "^0.10.0",
+        "@gridsuite/commons-ui": "^0.12.2",
```
sBouzols: It's not the scope of this PR. There is already a PR to update the comm-ui version used by gridexplore

sBouzols (changes requested):

sBouzols (approved):

PR #21, opened 2021-08-10
feat(gridexplore-app): use common-ui VirtualizedTable

sBouzols:
use common-ui VirtualizedTable instead of the repository one.


sBouzols:
[WIP] : need to release a new common-ui fixed version to remove console errors about default Styles when using VirtualizedTable

ChamseddineBhd (approved):

PR #22, opened 2021-08-10
fix dialog default enter behaviour + sort elements in alphabetical order in directory-tree & directory-content

ChamseddineBhd:
Signed-off-by: chamseddine <chamseddine.benhamed@ensi-uma.tn>

### DIFF
--- a/src/components/dialogs/delete-dialog.js
+++ b/src/components/dialogs/delete-dialog.js
@@ -36,7 +36,7 @@ const DeleteDialog = ({ open, onClose, onClick, title, message, error }) => {
 
     const handleKeyPressed = (event) => {
         if (open && event.key === 'Enter') {
-            handleClick();
+            handleClose();
         }
     };
 
--- a/src/components/directory-tree-view.js
+++ b/src/components/directory-tree-view.js
@@ -153,7 +153,6 @@ const DirectoryTreeView = ({ treeViewUID, mapData }) => {
 
     const handleOpenMenu = (event) => {
         setAnchorEl(event.currentTarget);
-        event.stopPropagation();
     };
 
     const handleCloseMenu = () => {
@@ -209,7 +208,6 @@ const DirectoryTreeView = ({ treeViewUID, mapData }) => {
     };
 
     function onContextMenu(e, nodeIds) {
-        e.preventDefault();
         handleSelect(nodeIds, false);
         handleOpenMenu(e);
     }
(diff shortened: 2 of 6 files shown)


src/components/dialogs/delete-dialog.js
```diff
@@ -36,7 +36,7 @@ const DeleteDialog = ({ open, onClose, onClick, title, message, error }) => {
 
     const handleKeyPressed = (event) => {
         if (open && event.key === 'Enter') {
-            handleClick();
+            handleClose();
```
sBouzols, line 39: Not sure here... I continue to think it should be handleClick even if its a delete dialog

src/components/tree-views-container.js
```diff
@@ -161,9 +161,11 @@ const TreeViewsContainer = () => {
         (children) => {
             dispatch(
                 setCurrentChildren(
-                    children.filter(
-                        (child) => child.type !== elementType.DIRECTORY
-                    )
+                    children
```
sBouzols, line 168: Not sure if it's the right place to sort... I think it's better to do it during rendering like in renderTree for TreeItems...
Maybe move this sorting code somewhere in directory-content render method ?

src/components/directory-tree-view.js
```diff
@@ -250,9 +248,13 @@ const DirectoryTreeView = ({ treeViewUID, mapData }) => {
                 }}
             >
                 {Array.isArray(node.children)
-                    ? node.children.map((child) =>
-                          renderTree(mapDataRef.current[child.elementUuid])
-                      )
+                    ? node.children
```
sBouzols: The sorting will not work for rootDirectories then (the first TreeItem in this recursive function)... we must add this sorting code in directory-container when we loop over rootdirectories to render TreeViews.

src/components/directory-tree-view.js
```diff
@@ -153,7 +153,6 @@ const DirectoryTreeView = ({ treeViewUID, mapData }) => {
 
     const handleOpenMenu = (event) => {
         setAnchorEl(event.currentTarget);
-        event.stopPropagation();
```
sBouzols, line 156: I don't see any differences then I imagine it's a good pratice to remove code which we don't understand behavior but maybe you could explain here why `event.stopPropagation()` and `e.preventDefault()` are useless

sBouzols (changes requested):

src/components/directory-tree-view.js
ChamseddineBhd: it's already added

src/components/directory-tree-view.js
ChamseddineBhd: because we have added  event.preventDefault() in app.js in a previous commit

sBouzols (approved):

PR #23, opened 2021-08-10
584 create root folder from empty space

sBouzols:


src/components/app.js
```diff
@@ -240,6 +240,9 @@ const App = () => {
                                     xs={12}
                                     sm={3}
                                     style={{
+                                        display: 'flex',
+                                        flexDirection: 'column',
```
sBouzols: That creates a bug and make a scrollbar appear... bad...

src/components/tree-views-container.js
```diff
@@ -47,6 +86,196 @@ const TreeViewsContainer = () => {
...
+            if (!nodeId) {
+                setShowCreateStudyMenuItem(false);
+                setShowCreateNewDirectoryMenuItem(false);
+            } /* open Contextual Menu in a TreeViewItem */ else {
+                setShowCreateStudyMenuItem(true);
+                setShowCreateNewDirectoryMenuItem(true);
+            }
+
+            if (nodeId !== selectedDirectoryRef.current) {
+                dispatch(setSelectedDirectory(nodeId));
```
ChamseddineBhd: i think this is already handled , isn't it?

src/components/tree-views-container.js
```diff
@@ -47,6 +86,196 @@ const TreeViewsContainer = () => {
...
+    const [
+        showCreateNewDirectoryMenuItem,
+        setShowCreateNewDirectoryMenuItem,
+    ] = React.useState(true);
+
+    /* User interactions */
+    const onContextMenu = useCallback(
+        (e, nodeId) => {
+            setMousePosition({
+                mouseX: e.clientX + 16,
```
ChamseddineBhd: maybe create constants to hold 16 and 4 values

src/components/app.js
```diff
@@ -245,7 +245,15 @@ const App = () => {
                                         marginTop: '20px',
                                     }}
                                 >
-                                    <TreeViewsContainer />
+                                    <div
```
ChamseddineBhd: you should do that in the TreeViewsContainer

ChamseddineBhd (changes requested):

src/components/tree-views-container.js
sBouzols: ok but I'm not sure we should allow to call `dispatch(setSelectedDirectory(null));` when the contextual menu is opened from the empty space, because it changes the content and breadcrumbs display and it should be made only if we create a new root directory effectively.
I propose to amend the code with this test only to test if nodeId is null or not:
`if (nodeId) `

src/components/tree-views-container.js
sBouzols: ok

src/components/tree-views-container.js
sBouzols: ok resolved by calling handleselect when rightclick on a treeitem before openning the contextual menu

src/components/directory-tree-view.js
```diff
@@ -88,59 +65,12 @@ const useStyles = makeStyles((theme) => ({
...
-        anchorOrigin={{
-            vertical: 'bottom',
-            horizontal: 'left',
-        }}
-        {...props}
-    />
-));
-
-const DirectoryTreeView = ({ treeViewUID, mapData }) => {
+const DirectoryTreeView = ({ treeViewUID, mapData, onContextMenu }) => {
```
ChamseddineBhd: treeViewUuid

src/components/directory-tree-view.js
```diff
@@ -224,9 +93,10 @@ const DirectoryTreeView = ({ treeViewUID, mapData }) => {
                 label={
                     <div
                         className={classes.treeItemLabelRoot}
-                        onContextMenu={(e) =>
-                            onContextMenu(e, node.elementUuid)
-                        }
+                        onContextMenu={(e) => {
```
ChamseddineBhd: extract this function plz

src/components/tree-views-container.js
```diff
@@ -289,16 +501,201 @@ const TreeViewsContainer = () => {
...
+                    flexDirection: 'column',
+                    height: '100%',
+                }}
+                onContextMenu={(e) => onContextMenu(e, null)}
+            >
+                {mapDataRef.current &&
+                    rootDirectories.map((rootDirectory) => (
+                        <DirectoryTreeView
+                            key={rootDirectory.elementUuid}
+                            treeViewUID={rootDirectory.elementUuid}
```
ChamseddineBhd: treeViewUuid

ChamseddineBhd (changes requested):

ChamseddineBhd (approved):

PR #24, opened 2021-08-11
feat(directory-content): open contextual menu on Mouse pointer position

sBouzols:


ChamseddineBhd (commented):
rebase branch please

src/components/directory-content.js
```diff
@@ -383,6 +379,10 @@ const DirectoryContent = () => {
                                         event.rowData.accessRights.private
                                     );
                                 }
+                                setMousePosition({
```
ChamseddineBhd, line 447: 16 and 4 values should be saved as a constants and should be shared with the same value of the other PR (root folder menu)

ChamseddineBhd (changes requested):

ChamseddineBhd (approved):

PR #32, opened 2021-08-18
fix(content,container): right click close contextual Menu

sBouzols:


laurgarn (approved):
Hopefully whitespace changes can be omitted in diff. Still, the 80 character per line produces noise in the diff.

AbdelHedhili (approved):

PR #33, opened 2021-08-20
contextual menu no longer select the directory

AbdelHedhili:


laurgarn (approved):
Mes pauvres yeux pas encore au point en React n'ont rien vu de suspect.

src/components/tree-views-container.js
```diff
@@ -149,6 +152,7 @@ const TreeViewsContainer = () => {
 
     const handleCloseMenu = () => {
         setAnchorEl(null);
+        setSelectedDirectoryByContextualMenu(null);
```
sBouzols: dispatch this

src/redux/reducer.js
```diff
@@ -42,6 +43,7 @@ const initialState = {
     computedLanguage: getLocalStorageComputedLanguage(),
     currentChildren: null,
     selectedDirectory: null,
+    selectedDirectoryByContextualMenu: null,
```
sBouzols: Not fan of this naming.  I would say the 'ByContextualMenu' is too much precise over a specific user action.
If tomorrow we delete a directory with the keyboard for example, there will be 'selectedDirectoryByContextualMenu=null'
I would prefer `activeDirectory` which is not equivalent as `selectedDirectory`, but actually the former `selectedDirectory` should become `activeDirectory` and the new one `selectedDirectory` but it could  be ambiguous...

sBouzols (changes requested):
We must keep the focus CSS style when opening the menu on a non selected directory I would say.

src/redux/reducer.js
AbdelHedhili: ok good idea

sBouzols (approved):

PR #39, opened 2021-09-02
Fix contingencies edition

FranckLecuyer:


src/components/dialogs/filters-contingency-dialog.js
```diff
@@ -69,7 +69,7 @@ const FiltersContingencyDialog = ({
     const [description, setDescription] = useState('');
 
     const handleClose = () => {
-        onClose();
+        handleCancel();
     };
 
     const handleCancel = () => {
```
sBouzols, line 75: In `handleCancel` I think you should call `onClose();` first then reset `NewFiltersContingency` and `BtnSaveListDisabled` to avoid graphic change when closing dialog

src/components/dialogs/script-contingency-dialog.js
```diff
@@ -71,7 +71,7 @@ const ScriptContingencyDialog = ({ listId, open, onClose, onError, title }) => {
     };
 
     const handleClose = () => {
-        onClose();
+        handleCancel();
     };
 
     const handleCancel = () => {
```
sBouzols, line 77: Same remark than before here for this `handleCancel`

sBouzols (commented):
Please fix another little issue

sBouzols (approved):

src/components/dialogs/filters-contingency-dialog.js
sBouzols: it doesn't fix this little issue.

PR #40, opened 2021-09-02
correct getLink Us625

laurgarn:
correct getLink to look for objectType in name, case insensitive


src/components/directory-content.js
```diff
@@ -298,8 +298,15 @@ const DirectoryContent = () => {
     function getLink(elementUuid, objectType) {
         let href = '#';
         if (appsAndUrls !== null) {
-            if (objectType === elementType.STUDY) {
-                href = appsAndUrls[1].url + '/studies/' + elementUuid;
+            let mayTs = appsAndUrls.filter(
+                (t) =>
+                    0 ===
+                    t.name.localeCompare(objectType, undefined, {
```
sBouzols: It's a problem to compare applications names with `objectTypes` I think, it works with `'Study'` but its a lucky strike.
If we had a `'Filter' `type to open with `action` app for example it doesn't work at all.

sBouzols (dismissed):

jonenst (approved):

PR #42, opened 2021-09-10
add filters gestion

AbdelHedhili:


src/components/dialogs/copy-to-script-dialog.js
```diff
@@ -40,13 +41,8 @@ const ContingencyCopyToScriptDialog = ({
...
     };
-
     const handleClick = () => {
-        newScriptFromFiltersContingencyList(
-            listId,
-            newNameValue,
-            selectedDirectory
-        )
+        onClick(listId, newNameValue, selectedDirectory)
             .then((response) => {})
```
sBouzols: keep it simple and manage the Promise outside of this component I think. onClick prop func should manage everything.
Don not call onClose here in the same spirit.

src/components/dialogs/replace-with-script-dialog.js
```diff
@@ -38,7 +39,7 @@ const ContingencyReplaceWithScriptDialog = ({
     };
 
     const handleClick = () => {
-        replaceFiltersWithScriptContingencyList(listId, selectedDirectory)
+        onClick(listId, selectedDirectory)
```
sBouzols: same remark than for the other dialog

src/components/directory-content.js
```diff
@@ -334,6 +383,16 @@ const DirectoryContent = () => {
         setActiveElement('');
     };
 
+    /**
+     * Filter script dialog: window status value for editing a filter script
+     */
+    const [openScriptDialog, setOpenScriptDialog] = React.useState(false);
```
sBouzols, line 449: I know it doesn't change a lot, but maybe we should keep states initialization at the begining of the component definition for readiness.

src/components/directory-content.js
```diff
@@ -525,13 +595,22 @@ const DirectoryContent = () => {
...
+                        res.map((e) => {
+                            metadata[e.id] = {
+                                name: e.name,
+                                filterType: e.type,
+                            };
+                            return e;
+                        });
+                        setChildrenMetadata(metadata);
+                    });
                 });
```
sBouzols: Maybe we should do something like that to avoid moving the set at each fetch addition :
```
.then(() => {
     setChildrenMetadata(metadata);
}
```

src/components/directory-content.js
```diff
@@ -525,13 +595,22 @@ const DirectoryContent = () => {
                                     name: e.name,
                                 };
                             });
-
-                            setChildrenMetadata(metadata);
                         }
                     );
+                })
+                .then(() => {
```
sBouzols: should be a then in the fetchContingencyListsInfos level no ? otherwise the metadata will be possibly partial

src/components/directory-content.js
```diff
@@ -625,6 +704,20 @@ const DirectoryContent = () => {
         );
     };
 
+    const allowsCopyFilterToScript = () => {
```
sBouzols, line 768: Seems like is the same allowing function than `allowsCopyContingencyToScript`, you could add a OR condition in `allowsCopyToScript`. No need to manage diferent allowing here for now at least.

src/components/directory-content.js
```diff
@@ -625,6 +704,20 @@ const DirectoryContent = () => {
...
+        let children = getSelectedChildren();
+        return children.length === 1 && children[0].type === elementType.FILTER;
+    };
+
+    const allowsReplaceFilterWithScript = () => {
```
sBouzols, line 773: same here

src/components/directory-content.js
```diff
@@ -744,6 +837,22 @@ const DirectoryContent = () => {
                                             setOpenScriptContingencyDialog(
                                                 true
                                             );
+                                        } else if (
```
sBouzols: Maybe we should add a `handleRowClick` function for readiness which take a `event.rowData` as parameter.

src/components/directory-content.js
```diff
@@ -865,6 +974,24 @@ const DirectoryContent = () => {
                                     )}
                                 </>
                             )}
+                            {allowsCopyFilterToScript() && (
```
sBouzols, line 985: We should add a function to build MenuItems because we mix allowance checks and types checks to show the good menu which is not clever.

src/translations/en.json
```diff
@@ -101,5 +103,27 @@
     "copyToScriptList": "Copy to script",
     "newNameList": "New list name",
     "copyToScript": "Copy to script",
-    "replaceWithScript": "Replace with script"
+    "replaceWithScript": "Replace with script",
+    "createNewFilter": "Créer un nouveau filtre",
```
sBouzols: french in here

src/translations/en.json
```diff
@@ -101,5 +103,27 @@
...
+    "VscConverterStations": "VSC converter stations",
+    "HvdcLines": "HVDC Lines",
+    "BusBarSections": "Busbar sections",
+    "range": "range",
+    "approx": "approximation",
+    "equality": "equality",
+    "Min": "Min",
+    "Max": "Max",
+
+    "FilterName": "Nom"
```
sBouzols: french

src/redux/reducer.js
```diff
@@ -51,6 +52,7 @@ const initialState = {
     cases: [],
     selectedCase: null,
     selectedFile: null,
+    filterList: null,
```
sBouzols: seems not used... I didn't find any references in this PR

src/components/dialogs/script-dialog.js
```diff
@@ -74,61 +77,91 @@ const ScriptContingencyDialog = ({ listId, open, onClose, onError, title }) => {
...
-                        setCurrentScriptContingency(data);
-                        setAceEditorContent(data.script);
-                        setName(data.name);
-                        setDescription(data.description);
-                    }
-                })
-                .catch((error) => {
-                    onError(error.message);
-                });
+            if (type === elementType.SCRIPT_CONTINGENCY_LIST) {
```
sBouzols, line 128: Not sure if its here that we have to manage a type and conditionally call APIs by type to get or save scripts... maybe we should manage this outside this component which is fed with a scriptdata, controlled prop and a onClick prop for save.

src/components/directory-content.js
sBouzols: Ok after a broader analysis, its an acceptable code for now, but we mix allowance and types check to build our menus and this is confusing.

sBouzols (changes requested):

src/components/directory-content.js
AbdelHedhili: i'll go against that, for each dialog we have states and function to handle that dialog at the same place it's as clear as possible + we already did like that for all the others dialogs

src/components/directory-content.js
AbdelHedhili: I don't get it, the set is filled little by little when we get the infos and the set is only done once at the end

src/components/directory-content.js
AbdelHedhili: refactoring this is not really the goal of my PR, I followed the way it is made. should probably make another pr if we want to change this

src/components/directory-content.js
sBouzols: Ok

src/components/directory-content.js
sBouzols: true

src/components/directory-content.js
sBouzols: my code proposal doesn't work

sBouzols (approved):

src/components/directory-content.js
```diff
@@ -916,28 +1046,60 @@ const DirectoryContent = () => {
...
+                currentName={activeElement ? activeElement.elementName : ''}
+                title={useIntl().formatMessage({ id: 'copyToScriptList' })}
+            />
+            <GenericFilterDialog
+                id={currentFilterId}
+                open={openGenericFilterDialog}
+                onClose={handleCloseGenericFilterDialog}
+                onError={handleError}
+                title={useIntl().formatMessage({ id: 'editFilter' })}
+            />
```
JacquesBorsenberger, line 1102: I'm not really fond of multiplying n dialogs for each element/action couple; a least not in an already big file.

JacquesBorsenberger (approved):

jonenst (approved):

PR #44, opened 2021-09-15
Fix contextual menu closure

etiennehomer:
Signed-off-by: HOMER Etienne <etiennehomer@gmail.com>

### DIFF
--- a/src/components/tree-views-container.js
+++ b/src/components/tree-views-container.js
@@ -578,7 +578,7 @@ const TreeViewsContainer = () => {
                     }
                 >
                     {/* Directories Menu */}
-                    {!showMenuFromEmptyZone && (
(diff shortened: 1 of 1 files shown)


src/components/tree-views-container.js
```diff
@@ -578,7 +578,7 @@ const TreeViewsContainer = () => {
                     }
                 >
                     {/* Directories Menu */}
-                    {!showMenuFromEmptyZone && (
```
sBouzols, line 581: strange fix. anchorEl is already used by the StyledMenu open prop, it should be ok.
The `showMenuFromEmptyZone` is suspicious (I added it :P)

sBouzols (approved):

PR #46, opened 2021-09-16
Limits re-renderings in TreeViewsContainer

laurgarn:
reduction of selectedDirectory/currentPath impact on not involved trees
use of currentPath selector only, in path bread crumb, directory treeview and directory content.


jonenst (changes requested):
bug when renaming a root element, the new name is not shown.

https://user-images.githubusercontent.com/89208/134167530-90bd18de-45ff-4df0-ab0b-1dae2ab9e877.mp4

jonenst:
Also breaks "CopyToScript" and "ReplaceWithScript" functionality (needs to get the element parent directory to copy into the same parent)

jonenst (approved):

PR #53, opened 2021-09-30
Fix bug on study creation error

etiennehomer:


src/components/create-study-form.js
```diff
@@ -273,12 +272,13 @@ export const CreateStudyForm = (props) => {
...
             activeDirectory
         ).then((res) => {
-            setCreateStudyErr('');
-            setStudyName('');
-            setStudyDescription('');
-            dispatch(removeSelectedFile());
-
-            if (!res.ok) {
+            if (res.ok) {
+                props.setOpen(false);
```
sBouzols: strange thing to edit props inside the component isn't it ?
If the 'open' state is outside this component, then we should add a `onClose` props to pass to this component and then call it nope ?

etiennehomer: done

src/components/create-contingency-list-form.js
```diff
@@ -75,7 +81,7 @@ export const CreateContingencyListForm = (props) => {
 
         let isPrivateContingencyList = contingencyListPrivacy === 'private';
 
-        props.setOpen(false);
+        onClose();
```
sBouzols: For homogenous code, you should move this call in a `if(res.ok)` in the Promise.then bellow.
Same thing that you did in create-study-form

sBouzols (changes requested):

src/components/tree-views-container.js
```diff
@@ -83,7 +81,7 @@ const TreeViewsContainer = () => {
     const activeDirectory = useSelector((state) => state.activeDirectory);
     const userId = useSelector((state) => state.user.profile.sub);
 
-    const currentEdit = useRef(null);
+    // const currentEdit = useRef(null);
```
sBouzols: remove this

src/components/create-filter-form.js
```diff
@@ -86,12 +89,13 @@ const CreateFilterDialog = ({
     title,
     customTextValidationBtn,
     customTextCancelBtn,
-    action,
+    activeDirectory,
```
sBouzols: this activeDirectory is in the redux/store, you could avoid passing it in props
`const activeDirectory = useSelector((state) => state.activeDirectory);`

src/components/create-filter-form.js
```diff
@@ -107,18 +111,38 @@ const CreateFilterDialog = ({
...
+                transient: true,
+            },
+            newNameList,
+            newListType,
+            filterPrivacy === 'private',
+            activeDirectory
+        ).then((res) => {
+            if (res.ok) {
+                onClose();
+                setCreateFilterErr('');
```
sBouzols: reset `setNewListName('')` and setNewListType('SCRIPT') as well maybe ?

sBouzols (changes requested):

src/components/tree-views-container.js
etiennehomer: done

src/components/create-filter-form.js
etiennehomer: done

src/components/create-filter-form.js
etiennehomer: i created a resetDialog for filter, contingencies and studies that reset all fields with the default values

sBouzols (approved):
fine code review
No tests

src/components/create-study-form.js
```diff
@@ -287,10 +298,14 @@ export const CreateStudyForm = (props) => {
...
                         .then((data) => {
-                            setCreateStudyErr(data.message);
+                            setCreateStudyErr(
+                                data.error + ' - ' + data.message
+                            );
                         })
                         .catch((error) => {
-                            setCreateStudyErr(error);
+                            setCreateStudyErr(
+                                error.error + ' - ' + error.message
```
jonenst: error.error is undefined

jonenst (changes requested):
error.error is undefined. For example when stop the gateway, you get:
"undefined - JSON.parse: unexpected character at line 1 column 1 of the JSON data"

jonenst (approved):

PR #55, opened 2021-09-30
fix(directory-content)

sBouzols:
Reset ids when closing filter/contingencies dialogs, isAllowed should return false by defaut if no children, add emptyMenu function to show the contextualMenu conditionnaly.

### DIFF
--- a/src/components/directory-content.js
+++ b/src/components/directory-content.js
@@ -338,6 +338,7 @@ const DirectoryContent = () => {
     const handleCloseFiltersContingency = () => {
         setOpenFiltersContingencyDialog(false);
         setActiveElement(null);
+        setCurrentFiltersContingencyListId(null);
     };
 
     const [
@@ -448,6 +449,7 @@ const DirectoryContent = () => {
     const handleCloseScriptContingency = () => {
         setOpenScriptContingencyDialog(false);
         setActiveElement(null);
+        setCurrentScriptContingencyListId(null);
     };
 
     /**
@@ -458,6 +460,7 @@ const DirectoryContent = () => {
     const handleCloseScriptDialog = () => {
         setOpenScriptDialog(false);
         setActiveElement('');
+        setCurrentScriptId(null);
     };
 
     const abbreviationFromUserName = (name) => {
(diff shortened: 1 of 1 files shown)


src/components/directory-content.js
```diff
@@ -817,6 +818,19 @@ const DirectoryContent = () => {
         );
     }
 
+    const emptyMenu = () => {
+        return !(
+            allowsRename() ||
+            allowsPublishability() ||
```
etiennehomer, line 824: The code of the isAllowed() method can be moved into AllowPublishability() and isAllowed() can be removed ?

src/components/directory-content.js
```diff
@@ -817,6 +818,19 @@ const DirectoryContent = () => {
...
+            allowsExport() ||
+            allowsDelete() ||
+            allowsCopyContingencyToScript() ||
+            allowsReplaceContingencyWithScript() ||
+            allowsCopyFilterToScript() ||
+            allowsReplaceFilterWithScript()
+        );
+    };
+
```
etiennehomer, line 833: In general, a lot a logic is implemented in those allowX methods. Maybe a back service could give us the allowed actions ?

src/components/directory-content.js
sBouzols: Not really I think, because its consistent with the isAllowed() function in directory-container. Actually this code needs a refactoring when we will add more functionalities in the directory-content toolbar, and a generalized management of actions on multi-selections.
We mainly check for selected items, types of items and privileges (user and access rights) for consistent allowed actions.

src/components/directory-content.js
sBouzols: It's an idea ! we could discuss of this new US with the team ! it allows to compute allowed actions only once from back service and not from the front.

etiennehomer: Yes, to be done later

src/components/directory-content.js
etiennehomer: Ok then

etiennehomer (approved):

jonenst (approved):

jonenst (approved):

PR #57, opened 2021-10-25
use explore server

etiennehomer:
Signed-off-by: HOMER Etienne <etiennehomer@gmail.com>


src/utils/rest-api.js
```diff
@@ -16,6 +16,8 @@ let PREFIX_CONFIG_NOTIFICATION_WS =
 let PREFIX_CONFIG_QUERIES = process.env.REACT_APP_API_GATEWAY + '/config';
 let PREFIX_DIRECTORY_SERVER_QUERIES =
     process.env.REACT_APP_API_GATEWAY + '/directory';
+let PREFIX_EXPLORE_SERVER_QUERIES =
```
sBouzols: should be `const` variable not `let`, same for others PREFIX_* (in another PR ?)

src/utils/rest-api.js
```diff
@@ -121,7 +123,7 @@ export function fetchDirectoryContent(directoryUuid) {
 export function deleteElement(elementUuid) {
```
sBouzols, line 123: Maybe we should give an order in this file to group REST API calls by service for readability, or even better separate into different files then import them in `rest-api.js`

sBouzols (changes requested):
fine code review
Maybe some refactoring to handle in another US ?

src/utils/rest-api.js
sBouzols: in another US / PR

FranckLecuyer (commented):
The functions : 

fetchStudiesInfos
fetchContingencyListsInfos
fetchFiltersInfos

in file rest-api.js

must call the explore-server (which then will call the directory-server and the study-server, actions-sever and filter-server)

FranckLecuyer (changes requested):
The replacement of a filters contingency list with a script contingency list is done in action-server, but the element type is not changed in the directory-server database.
Same problem with filters

src/components/directory-content.js
```diff
@@ -285,16 +288,21 @@ const DirectoryContent = () => {
...
                 setCurrentScriptContingencyListId(event.rowData.elementUuid);
                 setOpenScriptContingencyDialog(true);
-            } else if (event.rowData.type === elementType.SCRIPT) {
+            } else if (
+                event.rowData.type === elementType.FILTER &&
+                subtype === filterSubtype.SCRIPT
+            ) {
                 setCurrentScriptId(event.rowData.elementUuid);
                 setOpenScriptDialog(true);
             } else if (event.rowData.type === elementType.FILTER) {
```
sBouzols, line 302: `&&  subtype === filterSubtype.FILTER` is missing here.

src/components/directory-content.js
```diff
@@ -512,11 +520,38 @@ const DirectoryContent = () => {
         return href;
     }
 
+    function buildTypeWithSubtype(type, subtype) {
+        switch (type) {
+            case elementType.FILTER:
+                return subtype === filterSubtype.SCRIPT
+                    ? filterSubtype.SCRIPT
+                    : type;
```
sBouzols: replace `type` by `filterSubtype.FILTER` to be clearer

src/components/directory-content.js
```diff
@@ -512,11 +520,38 @@ const DirectoryContent = () => {
...
+            case elementType.CONTINGENCY_LIST:
+                return subtype === contingencyListSubtype.FILTERS
+                    ? contingencyListSubtype.FILTERS +
+                          '_' +
+                          elementType.CONTINGENCY_LIST
+                    : contingencyListSubtype.SCRIPT +
```
sBouzols: Not a fan of this syntax but it works... for now...

src/components/directory-content.js
```diff
@@ -535,17 +570,26 @@ const DirectoryContent = () => {
...
+            objectType === elementType.CONTINGENCY_LIST &&
+            objectSubtype === contingencyListSubtype.FILTERS
+        ) {
             return <PanToolIcon className={classes.icon} />;
+        } else if (
+            objectType === elementType.FILTER &&
+            objectSubtype === filterSubtype.SCRIPT
+        ) {
+            return <FilterIcon className={classes.icon} />;
         } else if (objectType === elementType.FILTER) {
```
sBouzols, line 573: `&&  objectSubtype === filterSubtype.FILTER` is missing

src/components/directory-content.js
```diff
@@ -563,10 +607,14 @@ const DirectoryContent = () => {
...
-                {childrenMetadata[elementUuid] && getElementIcon(objectType)}
+                {childrenMetadata[elementUuid] &&
+                    getElementIcon(
+                        objectType,
+                        childrenMetadata[elementUuid].subtype
+                    )}
                 {/* Name */}
                 {isMetadataLoading ? null : childrenMetadata[elementUuid] ? (
-                    <div>{childrenMetadata[elementUuid].name}</div>
+                    <div>{elementName}</div>
```
sBouzols: revert this, you should always prefer the name from the metadata I think

src/components/directory-content.js
```diff
@@ -774,29 +787,44 @@ const DirectoryContent = () => {
         let children = getSelectedChildren();
         return (
             children.length === 1 &&
-            children[0].type === elementType.FILTERS_CONTINGENCY_LIST
+            children[0].type === elementType.CONTINGENCY_LIST &&
+            childrenMetadata[children[0].elementUuid].subtype ===
```
sBouzols, line 778: maybe you should check if `childrenMetadata[children[0].elementUuid]` exists here first.
This is dirty code... OMG...

src/components/directory-content.js
```diff
@@ -774,29 +787,44 @@ const DirectoryContent = () => {
...
         );
     };
 
     const allowsReplaceContingencyWithScript = () => {
         let children = getSelectedChildren();
         return (
             children.length === 1 &&
-            children[0].type === elementType.FILTERS_CONTINGENCY_LIST &&
+            children[0].type === elementType.CONTINGENCY_LIST &&
+            childrenMetadata[children[0].elementUuid].subtype ===
```
sBouzols, line 789: same, must check if `childrenMetadata[children[0].elementUuid]` is OK first

src/components/directory-content.js
```diff
@@ -774,29 +787,44 @@ const DirectoryContent = () => {
...
     };
 
     const allowsCopyFilterToScript = () => {
         let children = getSelectedChildren();
-        return children.length === 1 && children[0].type === elementType.FILTER;
+        return (
+            children.length === 1 &&
+            children[0].type === elementType.FILTER &&
+            !(
+                childrenMetadata[children[0].elementUuid].subtype ===
```
sBouzols, line 802: same

src/components/directory-content.js
```diff
@@ -774,29 +787,44 @@ const DirectoryContent = () => {
...
+        );
     };
 
     const allowsReplaceFilterWithScript = () => {
         let children = getSelectedChildren();
         return (
             children.length === 1 &&
             children[0].type === elementType.FILTER &&
+            !(
+                childrenMetadata[children[0].elementUuid].subtype ===
```
sBouzols, line 815: same

src/components/directory-content.js
```diff
@@ -1087,15 +1115,17 @@ const DirectoryContent = () => {
                 onClose={handleCloseScriptContingency}
                 onError={handleError}
                 title={useIntl().formatMessage({ id: 'editContingencyList' })}
-                type={elementType.SCRIPT_CONTINGENCY_LIST}
+                type={elementType.CONTINGENCY_LIST}
+                subtype={contingencyListSubtype.SCRIPT}
```
sBouzols: I don't think you need this subtype props here. We check the subtype before setting `openScriptContingencyDialog` then it's obviously `contingencyListSubtype.SCRIPT`

src/components/directory-content.js
```diff
@@ -1087,15 +1115,17 @@ const DirectoryContent = () => {
...
             />
             <ScriptDialog
                 id={currentScriptId}
                 open={openScriptDialog}
                 onClose={handleCloseScriptDialog}
                 onError={handleError}
                 title={useIntl().formatMessage({ id: 'editFilterScript' })}
-                type={elementType.SCRIPT}
+                type={elementType.FILTER}
+                subtype={filterSubtype.SCRIPT}
```
sBouzols: same remark here

src/components/dialogs/script-dialog.js
```diff
@@ -51,8 +54,9 @@ const useStyles = makeStyles(() => ({
  * @param onError handle errors
  * @param title Title of the dialog
  * @param type Contingencies or filter
+ * @param subtype Element's subtype
  */
-const ScriptDialog = ({ id, open, onClose, onError, title, type }) => {
+const ScriptDialog = ({ id, open, onClose, onError, title, type, subtype }) => {
```
sBouzols: I don't thnik you need a subtype prop here, the main type is sufficient to manage conditions in this dialog.

sBouzols (changes requested):
Fine code Review
No tests

src/utils/rest-api.js
etiennehomer: Done in this one.

src/components/directory-content.js
etiennehomer: Done

src/components/directory-content.js
etiennehomer: done

src/components/directory-content.js
etiennehomer: Done.
Yes it's complicated... Maybe we should get the available menu actions from the back.

src/components/directory-content.js
etiennehomer: done

src/components/directory-content.js
etiennehomer: Actually, the subtype we get for the filter metadata is SCRIPT or LINE, LOAD, BATTERY etc. So we can't test if subtype === FILTER. It has to be changed in this US or not...

src/components/directory-content.js
etiennehomer: done

src/components/directory-content.js
etiennehomer: done

src/components/directory-content.js
etiennehomer: You're right. Done.

src/components/directory-content.js
etiennehomer: done

src/components/dialogs/script-dialog.js
etiennehomer: done

src/components/directory-content.js
sBouzols: It's not your dirty code :sweat_smile:

src/components/directory-content.js
etiennehomer: Changed was done in filter-server and in grid-explore

sBouzols (approved):

PR #60, opened 2021-11-15
Add ghost study

JacquesBorsenberger:


src/components/create-study-form.js
```diff
@@ -150,6 +154,11 @@ const UploadCase = () => {
     );
 };
 
+const ghostStudyKeyGenerator = (() => {
+    let key = 1;
```
laurgarn, line 158: An equivalent to a C function internal static ? Wow.

laurgarn (approved):

src/components/create-study-form.js
JacquesBorsenberger: i used the same mechanism as the one in merge orchestrator, generating an uuid is a bit overkill i think (need more dependency)

src/components/directory-content.js
```diff
@@ -120,6 +120,8 @@ const DirectoryContent = () => {
     const [selectedUuids, setSelectedUuids] = useState(new Set());
 
     const currentChildren = useSelector((state) => state.currentChildren);
+    const [displayedElements, setDisplayedElements] = useState([]);
```
sBouzols: I'm not sure you need another state here (displayedElements). You already have currentChildren and it creates confusion.
You may should complete `currentChildren` with matching (same directory) `uploadingStudies` in the `directory-container` component which synchronize data when browsing and everything will work as before with `currentChildren` in here.

sBouzols (changes requested):

src/components/directory-content.js
JacquesBorsenberger: current children are distant elements, and are selectable uploading ones are not, also when currentChildren changes we fetch associated metadata (useEffect),

src/components/create-study-form.js
```diff
@@ -285,31 +319,41 @@ export const CreateStudyForm = ({ open, onClose }) => {
...
+                    }
                 }
-            }
-        });
+            })
+            .catch((e) => {
+                studyCreationError(studyName, e.name + ' - ' + e.message);
+                dispatch(removeUploadingStudy(uploadingStudy));
+            });
+        dispatch(addUploadingStudy(uploadingStudy));
```
sBouzols, line 342: Could you move this before `createStudy` call ? I know it doesn't change anything with the Promise here which code is executed asynchronously, but it's more logical to `addUploadingStudy` first then `RemoveUploadingStudy`

src/components/tree-views-container.js
```diff
@@ -433,6 +436,33 @@ const TreeViewsContainer = () => {
...
+        if (updated) {
+            dispatch(
+                setCurrentChildren(
+                    Object.values(children).sort(function (a, b) {
+                        return a.elementName.localeCompare(b.elementName);
+                    })
+                )
+            );
+        }
+    }, [currentChildren, uploadingStudies, selectedDirectory, dispatch]);
```
sBouzols: Maybye use a `currentChildrenRef = useRef()` here to avoid the currentChildren dependency which could become tricky because you call `dispatch(setCurrentChildren(` in this useEffect.

src/components/create-study-form.js
JacquesBorsenberger: the call is not done, it is not uploading

src/components/directory-content.js
JacquesBorsenberger: made the change

src/components/tree-views-container.js
JacquesBorsenberger: done

src/components/create-study-form.js
sBouzols: True

src/components/directory-content.js
```diff
@@ -579,9 +579,21 @@ const DirectoryContent = () => {
...
+        const { elementName, uploading, elementUuid } = cellData.rowData;
+        const formatMessage = intl.formatMessage;
+        if (uploading)
+            return elementName + ' ' + formatMessage({ id: 'uploading' });
+        if (isMetadataLoading) return elementName;
+        if (childrenMetadata[elementUuid] == null)
+            return (
+                elementName + ' ' + formatMessage({ id: 'creationInProgress' })
+            );
+        return elementName;
```
sBouzols: should be `childrenMetadata[elementUuid].name` here to always prefer the name from metadata nope ?

sBouzols:
@JacquesBorsenberger and @laurgarn could you review my last changes please ?

src/components/directory-content.js
```diff
@@ -599,19 +611,18 @@ const DirectoryContent = () => {
...
+                {<div>{getDisplayedElementName(cellData)}</div>}
             </div>
         );
     };
 
     function toggleSelection(elementUuid) {
+        let element = currentChildren?.find(
+            (e) => e.elementUuid === elementUuid
+        );
+        if (element.uploading) {
```
laurgarn: No protection against absence from currentChildren ? (Does it make sense, either ? I have not checked)

src/components/directory-content.js
```diff
@@ -622,7 +633,11 @@ const DirectoryContent = () => {
     function toggleSelectAll() {
         if (selectedUuids.size === 0) {
             setSelectedUuids(
-                new Set(currentChildren.map((c) => c.elementUuid))
+                new Set(
+                    currentChildren
+                        .map((c) => c.elementUuid)
```
laurgarn: Are those flipped ? I would have put the filtering first and then the mapping for I don't trust an uuid has a uploading property.

src/components/directory-content.js
```diff
@@ -679,21 +694,29 @@ const DirectoryContent = () => {
...
-                            name: e.elementName,
-                            subtype: e.specificMetadata
-                                ? e.specificMetadata.type
-                                : null,
-                        };
+            let childrenToFetchElementsInfos = Object.values(currentChildren)
+                .filter((e) => !e.uploading)
+                .map((e) => e.elementUuid);
+            if (
+                childrenToFetchElementsInfos !== null &&
```
laurgarn: Object.values return value is an array, can not be null, even if passed null as parameter, as it would have thrown  Uncaught TypeError: can't convert null to object

laurgarn (changes requested):

laurgarn (approved):

sBouzols (approved):
Fine code review
Test on grid Explore study creation
Console messages Check OK
Miss creation in progress... after uploading (in another PR)
Must forbid click on study if in creation in progess state as well

PR #62, opened 2021-11-29
Add equipment type for form filter

etiennehomer:


src/components/dialogs/script-dialog.js
```diff
@@ -107,6 +107,7 @@ const ScriptDialog = ({ id, open, onClose, onError, title, type }) => {
                 description: description,
                 script: aceEditorContent,
                 type: filterSubtype.SCRIPT,
+                subtype: filterSubtype.SCRIPT,
```
AbdelHedhili: are both of these used ? It is weird that type and subtype use the same filterSubType enum. Should probably be contingencyListSubtype.SCRIPT no ?

AbdelHedhili (changes requested):

src/components/dialogs/script-dialog.js
etiennehomer: None of them were used actually. I removed those fields

niconoir (approved):

src/components/dialogs/script-dialog.js
```diff
@@ -129,8 +124,8 @@ const ScriptDialog = ({ id, open, onClose, onError, title, type }) => {
 
     const getCurrentScript = useCallback(
         (currentItemId) => {
-            if (type === elementType.CONTINGENCY_LIST) {
-                getContingencyList(contingencyListSubtype.SCRIPT, currentItemId)
+            if (type === ElementType.CONTINGENCY_LIST) {
+                getContingencyList(ContingencyListType.SCRIPT, currentItemId)
```
FranckLecuyer: Call getContingencyList only if currentItemId !== null (like in form-contingency-dialog.js) to avoid a bad request

FranckLecuyer (changes requested):

src/components/dialogs/script-dialog.js
etiennehomer: Done

src/components/dialogs/script-dialog.js
etiennehomer: Actually, type is used for script filter edition. I kept 
`type: FilterType.SCRIPT,`
It's clean and it works now

AbdelHedhili (approved):

FranckLecuyer (approved):

PR #66, opened 2021-12-10
US 721 show infinite loader when changing directory, until metadata is loaded

JacquesBorsenberger:


src/components/tree-views-container.js
```diff
@@ -453,14 +453,15 @@ const TreeViewsContainer = () => {
 
     const updateDirectoryTreeAndContent = useCallback(
         (nodeId) => {
+            dispatch(setCurrentChildren(undefined));
```
sBouzols: If a new study arrives in the current directory, all the content is updated. Maybe move this when we change directory.

src/components/directory-content.js
```diff
@@ -894,33 +906,35 @@ const DirectoryContent = () => {
...
-                            onClick={() => handleOpenDeleteElement()}
-                        >
-                            <DeleteIcon />
-                        </IconButton>
-                    )}
-                </Toolbar>
-                {selectedDirectory !== null &&
-                    currentChildren !== null &&
-                    currentChildren.length > 0 && (
+                {isMetadataLoading ||
```
sBouzols: Is removing condition on `selectedDirectory` dangerous ? What is shown when no directory is selected for exemple ?

src/components/directory-content.js
```diff
@@ -712,13 +721,16 @@ const DirectoryContent = () => {
                         });
                     })
                     .finally(() => {
-                        setChildrenMetadata(metadata);
-                        setIsMetadataLoading(false);
+                        // discarding request for older directory
+                        if (currentChildrenRef.current === currentChildren) {
```
sBouzols, line 736: This test has no effect, you just add `currentChildrenRef.current = currentChildren;` earlier, then there is no way this test could fail.

src/components/directory-content.js
```diff
@@ -69,6 +69,8 @@ import CopyToScriptDialog from './dialogs/copy-to-script-dialog';
 import { useSnackbar } from 'notistack';
 import GenericFilterDialog from './generic-filter';
 
+const circularProgressSize = 100;
```
sBouzols: Could you avoid using static size ?

sBouzols (changes requested):

src/components/tree-views-container.js
JacquesBorsenberger: modified,

src/components/directory-content.js
JacquesBorsenberger: nope, the function is asynchronous, this is a check to prevent setting erroneous data, you can test by adding a console.log in an else (throttle your network)

src/components/directory-content.js
JacquesBorsenberger: nope, nothing is shown

src/components/directory-content.js
JacquesBorsenberger: done

src/components/directory-content.js
sBouzols: Ok

sBouzols (approved):
Code review ok
Tests OK, will see with PO if showing loader during study creation is ok.

PR #67, opened 2021-12-14
fix element creation checks

AbdelHedhili:

### DIFF
--- a/src/components/create-contingency-list-form.js
+++ b/src/components/create-contingency-list-form.js
@@ -5,7 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 
 import Button from '@material-ui/core/Button';
 import TextField from '@material-ui/core/TextField';
@@ -19,13 +19,18 @@ import Alert from '@material-ui/lab/Alert';
 
 import { FormattedMessage, useIntl } from 'react-intl';
 
-import { createContingencyList } from '../utils/rest-api';
+import { createContingencyList, elementExists } from '../utils/rest-api';
 
 import { useSelector } from 'react-redux';
 import RadioGroup from '@material-ui/core/RadioGroup';
 import Radio from '@material-ui/core/Radio';
 import PropTypes from 'prop-types';
-import { ContingencyListType } from '../utils/elementType';
+import { ContingencyListType, ElementType } from '../utils/elementType';
+import CircularProgress from '@material-ui/core/CircularProgress';
+import { makeStyles } from '@material-ui/core/styles';
+import CheckIcon from '@material-ui/icons/Check';
+
+const useStyles = makeStyles(() => ({}));
 
 /**
  * Dialog to create a contingency
@@ -45,7 +50,15 @@ export const CreateContingencyListForm = ({ open, onClose }) => {
     const [createContingencyListErr, setCreateContingencyListErr] =
         React.useState('');
 
+    const [contingencyInvalid, setContingencyInvalid] = useState(false);
+    const [loadingCheckContingencyName, setLoadingCheckContingencyName] =
+        React.useState(false);
+    const [contingencyNameChecked, setContingencyNameChecked] =
+        React.useState(false);
+
+    const classes = useStyles();
     const intl = useIntl();
+    const timer = React.useRef();
 
     const activeDirectory = useSelector((state) => state.activeDirectory);
 
@@ -54,6 +67,8 @@ export const CreateContingencyListForm = ({ open, onClose }) => {
         setContingencyListDescription('');
         setContingencyListPrivacy('private');
         setContingencyListType(ContingencyListType.SCRIPT);
+        setContingencyNameChecked(false);
+        setLoadingCheckContingencyName(false);
         setCreateContingencyListErr('');
     };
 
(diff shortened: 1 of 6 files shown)


src/translations/fr.json
```diff
@@ -42,6 +42,9 @@
     "uploading": "téléversement en cours",
     "studyCreatingError": "Une erreur est survenue lors de la création de l'étude : {studyName}",
     "studyNameAlreadyUsed": "Ce nom d'étude est déjà utilisé",
+    "nameEmpty": "Le nom est vide",
+
+    "nameAlreadyUsed": "This name is already used",
```
sBouzols: must translate in french

src/components/create-filter-form.js
```diff
@@ -91,30 +95,82 @@ const CreateFilterDialog = ({
...
+
+    const handleFilterNameChanges = (name) => {
+        setNewListName(name);
+        setFilterNameChecked(false);
+        setLoadingCheckFilterName(true);
+
+        clearTimeout(timer.current);
+        timer.current = setTimeout(() => {
+            updateFilterFormState(name);
+        }, 700);
```
sBouzols, line 162: This hard value `700` deserves a variable somewhere

src/components/create-filter-form.js
```diff
@@ -91,30 +95,82 @@ const CreateFilterDialog = ({
...
         }
+        setLoadingCheckFilterName(false);
+    };
+
+    const handleFilterNameChanges = (name) => {
+        setNewListName(name);
+        setFilterNameChecked(false);
+        setLoadingCheckFilterName(true);
+
+        clearTimeout(timer.current);
```
sBouzols, line 159: Add comment here that this line clear Timers of previous calls to `handleFilterNameChanges` then this allow to call `updateFilterFormState` only on last keyboard input

src/components/create-filter-form.js
```diff
@@ -91,30 +95,82 @@ const CreateFilterDialog = ({
...
+                    intl.formatMessage({ id: 'nameEmpty' }),
+                    false
+                );
+            }
         } else {
-            setNewListName(name);
-            setDisableBtnSave(false);
+            setFilterFormState('', false);
         }
+        setLoadingCheckFilterName(false);
```
sBouzols: Maybe it should be in a .finally for the elementExists call and in some else as well, because here it's a quick request but with this actual code you could setLoadingCheckFilterName(false); before request done.

src/components/create-filter-form.js
```diff
@@ -164,13 +220,36 @@ const CreateFilterDialog = ({
...
-                            style={{ width: '100%' }}
+                            style={{ width: '90%' }}
                             defaultValue={''}
                             onChange={(event) =>
-                                onChangeInputName(event.target.value)
+                                handleFilterNameChanges(event.target.value)
                             }
                             label={inputLabelText}
                         />
+                        {loadingCheckFilterName && (
```
sBouzols: Maybe move this in a `renderFilterNameStatus` then you could be sure that there will be only one div with CircularProgress if loadingCheckFilterName else with CheckIcon if filterNameChecked or even a red cross if not.

src/components/create-filter-form.js
```diff
@@ -221,7 +300,7 @@ const CreateFilterDialog = ({
                     variant="outlined"
                     size="small"
                     onClick={handleSave}
-                    disabled={disableBtnSave}
+                    disabled={!filterNameChecked || filterInvalid}
```
sBouzols: `filterNameChecked` and `filterInvalid` represent the same value but inverted. I think we could use only one for now.

sBouzols (changes requested):
All my remarks in create-filter-fomr are valid for create study and create contingency.
No tests

src/components/create-contingency-list-form.js
```diff
@@ -67,7 +65,6 @@ export const CreateContingencyListForm = ({ open, onClose }) => {
         setContingencyListDescription('');
         setContingencyListPrivacy('private');
         setContingencyListType(ContingencyListType.SCRIPT);
```
sBouzols, line 67: reset `contingencyInvalid` here too

src/components/create-contingency-list-form.js
```diff
@@ -53,8 +53,6 @@ export const CreateContingencyListForm = ({ open, onClose }) => {
     const [contingencyInvalid, setContingencyInvalid] = useState(false);
```
sBouzols: maybe rename `contingencyInvalid` to `contingencyValid`. Must be coherent everywhere for a same state meaning because we have : 
* `contingencyInvalid`
* `filterInvalid`
but
* `studyValid`

src/components/create-filter-form.js
```diff
@@ -162,14 +165,12 @@ const CreateFilterDialog = ({
...
         setFilterInvalid(!isNameValid);
-        setFilterNameChecked(isNameValid);
     };
 
     const resetDialog = () => {
         setNewListName('');
         setNewListType(FilterType.SCRIPT);
         setFilterPrivacy('private');
-        setFilterNameChecked(false);
         setLoadingCheckFilterName(false);
```
sBouzols, line 174: reset filterInvalid here too

src/components/create-study-form.js
```diff
@@ -222,21 +220,43 @@ export const CreateStudyForm = ({ open, onClose }) => {
...
+        );
+    };
+
     const updateStudyFormState = (inputValue) => {
         if (inputValue !== '') {
             //If the name is not only white spaces
             if (inputValue.replace(/ /g, '') !== '') {
                 elementExists(activeDirectory, inputValue, ElementType.STUDY)
                     .then((data) => {
+                        console.log(data);
```
sBouzols: remove this console log

src/components/create-study-form.js
```diff
@@ -465,7 +470,7 @@ export const CreateStudyForm = ({ open, onClose }) => {
                     </Button>
                     <Button
                         onClick={() => handleCreateNewStudy()}
-                        disabled={!studyNameChecked || studyInvalid}
+                        disabled={studyName === '' || !studyNameValid}
```
sBouzols: + `|| loadingCheckStudyName` ?

src/components/create-study-form.js
```diff
@@ -398,33 +421,15 @@ export const CreateStudyForm = ({ open, onClose }) => {
                             margin="dense"
                             value={studyName}
                             type="text"
-                            error={studyInvalid}
+                            error={
```
sBouzols, line 422: Maybe we should add an error props for equivalent TextFileds component in create-filter-form.js and create-contingency-form.js 
Maybe other props as well, like autoFocus, margin, type, style, value ? and remove defaultValue elsewhere ?

src/components/create-study-form.js
```diff
@@ -222,21 +220,43 @@ export const CreateStudyForm = ({ open, onClose }) => {
         const name = e.target.value;
         setStudyName(name);
 
-        setStudyNameChecked(false);
         setLoadingCheckStudyName(true);
```
sBouzols, line 222: Maybe we should move this call at the beginning of `updateStudyFormState` ?

sBouzols (changes requested):
We should have homogenous code between create-*-form.js files

etiennehomer:
With the last changes, we lost the fact that the "create" button of the create study form is disable during the name check

src/components/create-study-form.js
AbdelHedhili: No it should not, if we do so (as the checks are really fast) we can't even see the loading sign. The point of it is to show that something is happening as soon as it starts to load

sBouzols: ok

sBouzols (approved):
Code review ok
Multiple tests done with grid-explore
* filter creation
* contingency list creation
* study creation

etiennehomer (approved):

PR #73, opened 2022-01-07
Move dialog components in dialog component folder, renaming, and import fixes

sBouzols:
Signed-off-by: sBouzols <sylvain.bouzols@gmail.com>


src/components/dialogs/create-contingency-list-dialog.js
```diff
@@ -250,7 +250,9 @@ export const CreateContingencyListForm = ({ open, onClose }) => {
                             }
                             type="text"
                             style={{ width: '90%' }}
-                            label=<FormattedMessage id="contingencyListName" />
+                            label={
```
etiennehomer: Braces are needed there ? 
There aren't used in such case in create-filter-dialog and create-study-dialog for example

sBouzols: I will remove this. I got a warning about this syntax in my IDE. I should fix it in another PR anyway.

etiennehomer (approved):

JacquesBorsenberger (approved):

PR #75, opened 2022-01-07
Handle error messages when creating and modifying elements

SlimaneAmar:


src/utils/rest-api.js
```diff
@@ -289,7 +289,11 @@ export function fetchElementsInfos(ids) {
     }).then((response) =>
         response.ok
             ? response.json()
-            : response.text().then((text) => Promise.reject(text))
+            : response
+                  .text()
+                  .then((text) =>
+                      Promise.reject(text ? text : response.statusText)
```
JacquesBorsenberger: can't be factorized ?  like function handleJSonResponse( response ) { 
if (response.ok) return response.json()
return response.text().then((text) => Promise.reject(text ? text : response.statusText))
}

SlimaneAmar: Yes an US was created for that

JacquesBorsenberger (changes requested):
factorize a little

src/utils/rest-api.js
SlimaneAmar: Done

JacquesBorsenberger (approved):

AbdelHedhili (approved):

PR #80, opened 2022-02-09
Fix problems with websocket always recreated

FranckLecuyer:


AbdelHedhili (approved):

src/components/tree-views-container.js
```diff
@@ -500,64 +501,67 @@ const TreeViewsContainer = () => {
...
     ]);
 
     useEffect(() => {
-        const ws = connectNotificationsUpdateStudies();
-        // Note: dispatch doesn't change
-        // cleanup at unmount event
-        return function () {
-            ws.close();
-        };
+        connectNotificationsUpdateStudies();
```
sBouzols: We never close the websocket then ?
We should have just a useEffect with empty dependency array parameter then create the ws at mounting and closing it at unmounting (return function).
The ws creation will be called only once without using useRef.
The problem comes when you start to add dependencies in the UseEffect then the behavior change completly.

sBouzols (changes requested):
Not ok the ws is never closed ?

src/components/tree-views-container.js
FranckLecuyer: Modifications done

FranckLecuyer: Tests have been done and the websocket is properly recreated when leaving/reopening the navigator, and also when gateway service is stopped and restarted.

src/components/tree-views-container.js
```diff
@@ -522,43 +541,32 @@ const TreeViewsContainer = () => {
...
                     }
-                    return;
-                }
 
-                if (directoryUuid) {
-                    if (directoryUuid === selectedDirectoryRef.current) {
-                        updateDirectoryTreeAndContent(directoryUuid);
-                    } else {
-                        updateDirectoryTree(directoryUuid);
+                    if (directoryUuid) {
```
sBouzols: Here if a directory which is not a root directory received an update notification you will not updateTree nor content.
You should do this I think (made in another PR :
```javascript
                if (isRootDirectory) {
                    updateRootDirectories();
                    if (
                        selectedDirectoryRef.current != null && // nothing to do if nothing already selected
                        notificationTypeHeader ===
                            notificationType.DELETE_DIRECTORY &&
                        selectedDirectoryRef.current.elementUuid ===
                            directoryUuid
                    ) {
                        dispatch(setSelectedDirectory(null));
                    }
                    return;
                }

                if (directoryUuid) {
                    // Remark : It could be a Uuid of a rootDirectory if we need to update it because its content update
                    // if dir is actually selected then call updateDirectoryTreeAndContent of this dir
                    // else expanded or not then updateDirectoryTree
                    if (selectedDirectoryRef.current != null) {
                        if (
                            directoryUuid ===
                            selectedDirectoryRef.current.elementUuid
                        ) {
                            updateDirectoryTreeAndContent(directoryUuid);
                            return; // break here
                        }
                    }
                    updateDirectoryTree(directoryUuid);
                }
```
I added a check on `if (selectedDirectoryRef.current != null) {` too, because renaming a rootDirectory with right click when nothing is selected caused a crash otherwise ;)

sBouzols (changes requested):

src/components/tree-views-container.js
FranckLecuyer: Corrected

sBouzols (approved):
Ok

PR #81, opened 2022-03-09
Us 800 import case in grid explore

JacquesBorsenberger:


laurgarn (approved):

src/utils/rest-api.js
```diff
@@ -348,6 +348,28 @@ export function createStudy(
     }
 }
 
+export function createCase({ name, description, file, parentDirectoryUuid }) {
+    console.info('Creating a new case...');
+    let urlSearchParams = new URLSearchParams();
+    urlSearchParams.append('description', description);
```
sBouzols: We may call `encodeURIComponent` over `description` at least I think.

src/utils/messages.js
```diff
@@ -46,3 +47,24 @@ export function displayErrorMessageWithSnackbar({
         style: { whiteSpace: 'pre-line' },
     });
 }
+
+export const useSnackbarMessage = () => {
+    const intlRef = useIntlRef();
+    const { enqueueSnackbar } = useSnackbar();
+    let message;
+    message = useCallback(
+        (msg, headerMessage, headerValues) =>
```
sBouzols: Rename `headerMessage` by `headerMessageId`

sBouzols: Maybe `msgId` as well ? and keep interntionalization process in here ?

src/components/directory-content.js
```diff
@@ -357,6 +358,8 @@ const DirectoryContent = () => {
             } else if (objectSubtype === FilterType.FORM) {
                 return <FilterListIcon className={classes.icon} />;
             }
+        } else if (objectType === ElementType.CASE) {
+            return <PhotoIcon />;
```
sBouzols: Add `className` as well here for icon size and all...

src/components/dialogs/upload-case.js
```diff
@@ -0,0 +1,56 @@
...
+import React from 'react';
+
+export const UploadCase = () => {
+    const dispatch = useDispatch();
+    const selectedFile = useSelector((state) => state.selectedFile);
+
+    const handleFileUpload = (e) => {
+        e.preventDefault();
+        let files = e.target.files;
+        dispatch(selectFile(files[0]));
```
sBouzols, line 21: We don't manage multiple imports ? Maybe it's not that hard... take a look...

src/components/dialogs/upload-case.js
```diff
@@ -0,0 +1,56 @@
...
+    };
+
+    return (
+        <table>
```
sBouzols, line 25: We may have MUI components which realize this instead of plain HTML nodes...

src/components/dialogs/field-hook.js
```diff
@@ -0,0 +1,183 @@
...
+        setChecking(true);
+        setError(undefined);
+        timer.current = setTimeout(() => updateValidity(name), 700);
+    }, [name, updateValidity]);
+
+    useEffect(() => {
+        setError(undefined);
+        timer.current = undefined;
+        setChecking(undefined);
+        setAdornment(undefined);
```
sBouzols, line 172: Do we need to reset adornment ? Seems a lot static for me...

src/components/dialogs/field-hook.js
```diff
@@ -0,0 +1,183 @@
...
+
+const makeAdornmentEndIcon = (content) => {
+    return {
+        endAdornment: (
+            <InputAdornment
+                position="end"
+                // hack to circumviate centering of adornment
+                // when TextField has variant 'filled' with 'end' position
+                // : classes.adornRightOther
+                //className={classes.adornRightFilled}
```
sBouzols: Where is this JSS class `adornRightFilled` ? Remove this comment ?

src/components/dialogs/field-hook.js
```diff
@@ -0,0 +1,183 @@
...
+            <TextField
+                key={id}
+                size="small"
+                margin="dense"
+                id={id}
+                label={<FormattedMessage id={label} />}
+                value={value}
+                style={{ width: '100%' }}
+                onChange={handleChangeValue}
+                autoFocus={autoFocus}
```
sBouzols: This seems strange. We should let the custom hook user set the autoFocus props to the desired field. If we have multiple `useTextValue` in a dialog for exemple, which one will be focused ?

src/components/dialogs/upload-case.js
sBouzols: Ok I see it's a code move... Will see later then

sBouzols (commented):
some comments after code review, will test now

src/components/dialogs/upload-case.js
JacquesBorsenberger: how do we manage name ?

src/components/dialogs/field-hook.js
JacquesBorsenberger: adornment is the check when the name is valid

src/utils/messages.js
JacquesBorsenberger: msg is not an identifier but a message

src/components/dialogs/field-hook.js
JacquesBorsenberger: the one we put autofocus:true on the useTextValue :)

src/utils/rest-api.js
JacquesBorsenberger: done

src/utils/messages.js
JacquesBorsenberger: done for headerMessage

src/components/directory-content.js
JacquesBorsenberger: done

src/components/dialogs/field-hook.js
JacquesBorsenberger: done

src/components/dialogs/upload-case.js
sBouzols: Ha you'r right... will see later

src/components/dialogs/field-hook.js
sBouzols: HO I see... it makes confusion with adornment which we use for Unit in some forms textFields... Ok then

sBouzols (changes requested):
Tests Done: 
desc in case importation is not reset
case file path in case importation AND study creation is not reset (idea ?)

sBouzols (approved):
OK

PR #83, opened 2022-03-22
Material UI v5 migration

sBouzols:


sBouzols:
waiting to use commons-ui version with MUI V5 to merge this PR
That's the reason why Github CI don't pass. The current commons-ui used here (for now) needs @material-ui/* packages which were removed.

src/components/dialogs/filters-editor.js
```diff
@@ -27,7 +27,7 @@ const useStyles = makeStyles(() => ({
 
 const CustomNativeSelect = withStyles((theme) => ({
     select: {
-        color: theme.palette.type === 'light' ? '#000 !important' : '#fff',
+        color: theme.palette.mode === 'light' ? '#000 !important' : '#fff',
```
laurgarn, line 30: Has this been corrected through the upgrade, by any chance ?

src/components/dialogs/filters.js
```diff
@@ -54,7 +53,7 @@ export const CountriesSelection = ({ initialValue, onChange, disabled }) => {
             onChange={(oldVal, newVal) => onChange(newVal)}
             options={Object.keys(countriesList.object())}
             getOptionLabel={(code) => countriesList.get(code)}
-            renderInput={(props) => <TextField {...props} variant="outlined" />}
+            renderInput={(props) => <TextField {...props} />}
```
laurgarn, line 56: Could this be a return to the default ?

laurgarn (approved):

src/components/dialogs/filters.js
sBouzols: this is the default value for `TextField` `variant` prop then it's useless to specify it in V5

src/components/dialogs/filters-editor.js
sBouzols: I don't even know why we use a CustomNativeSelect here ? 
not in the scope I would say

jonenst (approved):

src/components/app.test.js
```diff
@@ -39,12 +44,16 @@ it('renders', async () => {
...
                     <Provider store={store}>
-                        <ThemeProvider theme={createTheme({})}>
-                            <SnackbarProvider hideIconVariant={false}>
-                                <CssBaseline />
-                                <App />
-                            </SnackbarProvider>
-                        </ThemeProvider>
+                        <StyledEngineProvider injectFirst>
+                            <ThemeProvider
+                                theme={createTheme(adaptV4Theme({}))}
```
niconoir: Can't `adaptV4Theme` be removed here too ?

niconoir (commented):
Still a call to adaptV4Theme which can be removed.

src/components/app.test.js
sBouzols: removed

PR #85, opened 2022-03-25
Ts 830 protect agianst null is private

laurgarn:


src/components/dialogs/access-rights-dialog.js
```diff
@@ -47,9 +47,7 @@ const AccessRightsDialog = ({
     const classes = useStyles();
 
     const [loading, setLoading] = React.useState(false);
-    const [selected, setSelected] = React.useState(
-        isPrivate !== undefined ? isPrivate.toString() : null
-    );
+    const [selected, setSelected] = React.useState(isPrivate?.toString());
```
AbdelHedhili: We don't want the value to be set to undefined if isPrivate is null

AbdelHedhili (changes requested):

AbdelHedhili (approved):

PR #87, opened 2022-03-31
Create study from imported case

Meklo:


Meklo:
I just discovered a bug: if you opened create study dialog through a case, when you right-click on it, its display switches to the other mode. I will push a fix as soon as possible.

src/components/menus/content-contextual-menu.js
```diff
@@ -17,6 +18,7 @@ import ScriptDialog from '../dialogs/script-dialog';
 import ReplaceWithScriptDialog from '../dialogs/replace-with-script-dialog';
 import CopyToScriptDialog from '../dialogs/copy-to-script-dialog';
 import GenericFilterDialog from '../dialogs/generic-filter-dialog';
+import CreateStudyForm from '../dialogs/create-study-dialog';
```
FranckLecuyer: Changing name of imported component is confusing

src/components/menus/content-contextual-menu.js
```diff
@@ -337,6 +346,16 @@ const ContentContextualMenu = (props) => {
             });
         }
 
+        if (allowsCreateNewStudyFromCase()) {
+            menuItems.push({
+                messageDescriptorId: 'createNewStudyFromImportedCase',
+                callback: () => {
+                    handleOpenDialog(DialogsId.ADD_NEW_STUDY);
```
FranckLecuyer, line 376: Missing ADD_NEW_STUDY in DialogsId

src/components/menus/content-contextual-menu.js
```diff
@@ -545,6 +564,13 @@ const ContentContextualMenu = (props) => {
                 onError={handleLastError}
                 title={useIntl().formatMessage({ id: 'editFilter' })}
             />
+
+            <CreateStudyForm
+                open={openDialog === DialogsId.ADD_NEW_STUDY}
+                onClose={handleCloseDialog}
+                providedCase={activeElement ? activeElement : undefined}
```
FranckLecuyer: providedCase={activeElement} should be sufficient ?

src/components/dialogs/create-study-dialog.js
```diff
@@ -365,8 +422,50 @@ export const CreateStudyDialog = ({ open, onClose }) => {
...
+                                                    paddingRight: '30px',
+                                                }}
+                                                color="primary"
+                                                component="label"
+                                            >
+                                                <FormattedMessage id="showSelectDirectoryDialog" />
+                                            </Button>
+                                        </th>
+                                        <th>
+                                            <p>{activeDirectoryName}</p>
```
FranckLecuyer: Maybe display the full directory path instead of just the name ?

src/components/dialogs/create-study-dialog.js
```diff
@@ -365,8 +422,50 @@ export const CreateStudyDialog = ({ open, onClose }) => {
...
                         label={<FormattedMessage id="studyDescription" />}
                     />
-                    {caseExist && <SelectCase />}
-                    {!caseExist && <UploadCase />}
+                    {!providedCase ? (
+                        (caseExist && <SelectCase />) ||
+                        (!caseExist && <UploadCase />)
+                    ) : (
+                        <>
+                            <table>
```
FranckLecuyer: Why using a table here ?

FranckLecuyer (changes requested):

src/components/dialogs/create-study-dialog.js
Meklo: I wanted to reproduce the same display as in the UploadCase component, so for the sake of homogeneity I inserted the same HTML structure.
I'll see how to simplify this, should I update UploadCase too?

src/components/dialogs/create-study-dialog.js
Meklo: I'll try to implement this

src/components/menus/content-contextual-menu.js
Meklo: Indeed, it will be updated

src/components/menus/content-contextual-menu.js
Meklo: True, it will be updated

src/components/dialogs/create-study-dialog.js
FranckLecuyer: If it's possible, maybe you can reuse the UploadCase component and change this component to use the material iu Table component ?

Meklo: I don't think it will do the trick since i was inspired by UploadCase only for its layout, its logic is not what i'm looking for here.
For now I've recreated the same view by replacing `<table>` simply by a `<div>` and adding a few margins.

FranckLecuyer (changes requested):
Just correct the 'CaseValidate' english translation in en.json file, which is incorrect

FranckLecuyer (approved):

jonenst (approved):

PR #91, opened 2022-04-13
Fix dialog right click behaviour

Meklo:
In order to solve the problem it was necessary to block event bubbling triggered from right-click events: onContextMenu and onMouseDown. 
They triggered undesired behaviour like showing contextual menu or reseting variables while a dialog box was open.
The first step consisted in swapping onMouseDownCapture to onMouseDown in DirectoryContent to reverse the direction of event propagation.
Next it was needed to override onContextMenu and onMouseDown in each dialog box to stop event propagation to DirectoryContent.
To avoid code duplication I created a component encapsulating dialog boxes related to the contextual menu: ContextualMenuDialog.
ContextualMenuDialog has been placed in all components concerned by the described anomaly.


JacquesBorsenberger (approved):
not sure fore the name but I don't have better to propose.

src/components/dialogs/contextual-menu-dialog.js
```diff
@@ -0,0 +1,33 @@
...
+            onClose={handleClose}
+            aria-labelledby={ariaLabel}
+            onKeyPress={handleKeyPressed}
+            //Being in a dialog box, we don't want the right click event or oncontextmenu to bubble to the parent component
+            onContextMenu={(e) => e.stopPropagation()}
+            onMouseDown={(e) => {
+                if (e.button === constants.MOUSE_EVENT_RIGHT_BUTTON) {
+                    e.stopPropagation();
+                }
+            }}
```
jonenst: use ...props to propagate all props ?

jonenst (approved):

PR #94, opened 2022-04-15
fix:restore create study dialog validation

Meklo:


klesaulnier (approved):
Tested with different target directories, with provided case or not : OK

PR #98, opened 2022-05-05
Us 886 multi del ghost n alerts

laurgarn:


JacquesBorsenberger (approved):

src/components/directory-content.js
```diff
@@ -99,7 +99,7 @@ const DirectoryContent = () => {
     const dispatch = useDispatch();
 
     const [childrenMetadata, setChildrenMetadata] = useState({});
-    const [isAllDataPresent, setIsAllDataPresent] = useState(false);
+    const isAllDataPresent = true;
```
niconoir: Since isAllDataPresent is always true (it seems it is never modified), maybe it is possible to completely remove it ?

niconoir (changes requested):

src/components/directory-content.js
laurgarn: You are right. I am removing it.

niconoir (approved):

PR #107, opened 2022-06-07
fix: big progress appearing when first element in directory takes tim…

laurgarn:
…e to create

### DIFF
--- a/src/components/directory-content.js
+++ b/src/components/directory-content.js
@@ -496,9 +496,6 @@ const DirectoryContent = () => {
                                 : null,
                         };
                     });
-                    if (res.length === 0) {
-                        metadata['got'] = 'timeout';
-                    }
                 })
                 .catch((reason) => {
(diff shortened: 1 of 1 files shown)


src/components/directory-content.js
```diff
@@ -469,35 +471,47 @@ const DirectoryContent = () => {
...
+                        metadata[e.elementUuid] = {
+                            name: e.elementName,
+                            subtype: e.specificMetadata
+                                ? e.specificMetadata.type
+                                : null,
+                        };
                     });
-            }
+                    if (res.length === 0) {
+                        metadata['got'] = 'timeout';
```
klesaulnier: is it useful somewhere ?

klesaulnier (approved):
Tested : OK
Big progess icon disappears quickly when directory is loaded even if 1st element is still loading

src/components/directory-content.js
laurgarn: Well spot. It was a workaround. I remove it.

PR #108, opened 2022-06-07
add status to situation

bendaamerahm:
add  (loader + status)  to situation


src/components/dialogs/create-case-dialog.js
```diff
@@ -104,7 +115,7 @@ export function CreateCaseDialog({ onClose, open }) {
                     <FormattedMessage id="cancel" />
                 </Button>
                 <Button
-                    onClick={() => handleCreateNewStudy()}
+                    onClick={() => handleCreateNewCase()}
```
JacquesBorsenberger: just handleCreateNewCase to avoid creating new callback at each renders ;)

src/components/directory-content.js
```diff
@@ -382,7 +382,8 @@ const DirectoryContent = () => {
             <div className={classes.cell}>
                 {/*  Icon */}
                 {!childrenMetadata[elementUuid] &&
-                    objectType === ElementType.STUDY && (
+                    (objectType === ElementType.STUDY ||
```
JacquesBorsenberger: maybe use a function to be more meaningful, if we add a third type it will become really ugly

src/components/tree-views-container.js
```diff
@@ -194,6 +194,7 @@ const TreeViewsContainer = () => {
     const activeDirectory = useSelector((state) => state.activeDirectory);
 
     const uploadingStudies = useSelector((state) => state.uploadingStudies);
```
JacquesBorsenberger: maybe reusing the existing system to be more generic, like  uploadingElement ? instead of creating the same system twice ?

src/components/tree-views-container.js
```diff
@@ -347,16 +348,26 @@ const TreeViewsContainer = () => {
...
                 (e) =>
                     e.directory === selectedDirectoryRef.current.elementUuid &&
                     current[e.elementName] === undefined
             );
 
-            if (toMerge != null && toMerge.length > 0) {
-                return [...current, ...toMerge].sort(function (a, b) {
-                    return a.elementName.localeCompare(b.elementName);
-                });
+            if (
```
JacquesBorsenberger: as we are copy/paste we are multiplying similar lines

src/components/directory-content.js
bendaamerahm: yes,ill put it in a function

src/components/tree-views-container.js
bendaamerahm: ill do the refactor to use one object as you propose (uploadingElement)

JacquesBorsenberger (approved):
:) I don't have better name for elementTypeVerification so it is good for me

sBouzols (approved):
Code Review Ok
Tests done
Warning check Ok

PR #113, opened 2022-06-23
add 2 scrollbars to the TreeViewsContainer Grid

dbraquart:
To explore and navigate into a huge or deep tree, scrollbars have been added.
Impacts:
- there is no more truncated names (ending with "...")
- the private/lock icon is now at the left of the directory name

Also changed: user selection in the tree view is now disabled.

### DIFF
--- a/src/components/directory-tree-view.js
+++ b/src/components/directory-tree-view.js
@@ -27,6 +27,7 @@ const useStyles = makeStyles((theme) => ({
         padding: theme.spacing(0.5),
     },
     treeItemRoot: {
+        userSelect: 'none',
         '&:focus > $treeItemContent $treeItemLabel, .focused': {
             borderRadius: theme.spacing(2),
             backgroundColor: theme.row.primary,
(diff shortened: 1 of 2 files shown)


laurgarn (approved):

src/components/directory-tree-view.js
```diff
@@ -27,6 +27,7 @@ const useStyles = makeStyles((theme) => ({
         padding: theme.spacing(0.5),
     },
     treeItemRoot: {
+        userSelect: 'none',
```
klesaulnier, line 30: why do we need this ?

klesaulnier (approved):
Test : OK
Review : OK except 1 confusion on a css property

src/components/directory-tree-view.js
dbraquart: userSelect=none is to disable the ability to select text from the tree. This is useless, and it happens often by mistake. 
(validated with PO)
select-tree-view-orange

dbraquart:
Add scrollbars in the tree view panel

PR #118, opened 2022-06-28
Initial commit

Meklo:


laurgarn (approved):
Message for initial commit could be more informative ;-)
Although branch name tells already

Meklo:
> Message for initial commit could be more informative ;-) Although branch name tells already

Indeed, I wasn't aware the commit message on main is based on the first commit message by default ^^'

PR #124, opened 2022-07-06
Snackbar error is now displayed when element duplication fails

klesaulnier:
Signed-off-by: LE SAULNIER Kevin <kevin.lesaulnier@rte-france.com>

### DIFF
--- a/src/components/menus/content-contextual-menu.js
+++ b/src/components/menus/content-contextual-menu.js
@@ -98,87 +98,99 @@ const ContentContextualMenu = (props) => {
                 selectedDirectory.elementUuid,
                 activeElement.elementName,
                 activeElement.type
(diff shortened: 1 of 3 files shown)


Meklo (changes requested):
If I'm not mistaken your addition handles the case where getNameCandidate fails, I tested it, works like a charm.
However I noticed if an error occur during the duplication itself, its message is displayed but the snackbar has no title because it's handled this way :
```
                                    .catch((message) => {
                                        handleLastError(message);
                                    });
```
                                    
Maybe add a function which would handle all errors related to the duplication in order to always format the snackbar with the duplication error title duplicateElementFailure ?

FranckLecuyer (approved):

Meklo (approved):

PR #129, opened 2022-07-11
Fix bug status create study from case

bendaamerahm:
Fix bug status create study from case

Signed-off-by: bendaamerahm <ahmed.bendaamer@rte-france.com>

### DIFF
--- a/src/components/dialogs/upload-case.js
+++ b/src/components/dialogs/upload-case.js
@@ -18,6 +18,7 @@ export const UploadCase = () => {
     const handleFileUpload = (e) => {
         e.preventDefault();
         let files = e.target.files;
+        if (files.size === 0) dispatch(selectFile(null));
         dispatch(selectFile(files[0]));
     };
 
@@ -42,7 +43,7 @@ export const UploadCase = () => {
                     </th>
                     <th>
                         <p>
-                            {selectedFile === null ? (
+                            {selectedFile?.name === undefined ? (
                                 <FormattedMessage id="uploadMessage" />
                             ) : (
                                 selectedFile.name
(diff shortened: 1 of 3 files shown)


src/components/dialogs/upload-case.js
```diff
@@ -18,6 +18,7 @@ export const UploadCase = () => {
     const handleFileUpload = (e) => {
         e.preventDefault();
         let files = e.target.files;
+        if (files.size === 0) dispatch(selectFile(null));
         dispatch(selectFile(files[0]));
```
sBouzols: then you don't need the second dispatch call ? you must return before this line I thnik if `files.size === 0`

src/components/tree-views-container.js
```diff
@@ -352,11 +352,22 @@ const TreeViewsContainer = () => {
                     e.directory === selectedDirectoryRef.current.elementUuid &&
                     current[e.elementName] === undefined
             );
-
             if (elementsToMerge != null && elementsToMerge.length > 0) {
-                return [...current, ...elementsToMerge].sort(function (a, b) {
-                    return a.elementName.localeCompare(b.elementName);
-                });
+                const filtredCurrentElements = current.filter(
```
sBouzols, line 358: Add a comment
// We need to filter current array of elements in elementsToMerge to avoid duplicates in the directoryContent component.
// An uploading element doesn't have an elementUuid yet, then we filter on element Name and type.

sBouzols (approved):
Code Review -> minor changes requested
Test OK

src/components/dialogs/upload-case.js
bendaamerahm: nice catch ill add the else ,my bad

YenguiSeddik (approved):
Code review: OK
Test: OK

PR #140, opened 2022-08-16
HTTP404 -> HTTP204

EstherDarkish:
Uses a HTTP204 return code instead of HTTP404 when checking if a directory or element exists.

### DIFF
--- a/src/utils/rest-api.js
+++ b/src/utils/rest-api.js
@@ -422,11 +422,10 @@ export function elementExists(directoryUuid, elementName, type) {
     console.debug(existsElementUrl);
     return backendFetch(existsElementUrl, { method: 'head' }).then(
         (response) => {
-            return response.ok
-                ? true
-                : response.status === 404
-                ? false
-                : Promise.reject(response.statusText);
+            if (response.ok) {
+                return response.status !== 204; // HTTP 204 : No-content
+            }
+            return Promise.reject(response.statusText);
         }
     );
 }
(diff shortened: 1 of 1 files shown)


laurgarn (approved):

bendaamerahm (approved):

src/utils/rest-api.js
```diff
@@ -422,11 +422,10 @@ export function elementExists(directoryUuid, elementName, type) {
...
         (response) => {
-            return response.ok
-                ? true
-                : response.status === 404
-                ? false
-                : Promise.reject(response.statusText);
+            if (response.ok) {
+                return response.status !== 204; // HTTP 204 : No-content
+            }
+            return Promise.reject(response.statusText);
```
SlimaneAmar, line 428: return response.ok ?
? response.status !== 204
: response.text().then((text) => Promise.reject(text)) :

SlimaneAmar (approved):

src/utils/rest-api.js
EstherDarkish: I don't like mixing ternary operators and multiple return types. When reading the code, it can be easy to overlook the fact that we could return three different things : true, false, and reject.

I think it is easier to read something like this : 
`return condition ? value : value;`
than something like this : 
`return condition ? conditionalValue : value;`
In the second example, it is easy to overlook the `conditionalValue` and not realize it could lead to multiple and different return values.

By using a simple if clause, it is easier to differenciate between the conditional returns and the reject one.

But if you want the change, just say it and i'll change it.

SlimaneAmar: Right

PR #149, opened 2022-10-07
Update form filter creation

dbraquart:


src/components/dialogs/filters.js
```diff
@@ -45,38 +70,53 @@ export const CountriesSelection = ({ initialValue, onChange, disabled }) => {
...
-                        size={'small'}
-                        label={countriesList.get(code)}
-                        {...getTagsProps({ index })}
-                    />
-                ))
-            }
-        />
+        <>
+            <FormControl fullWidth margin="dense">
+                <Autocomplete
```
laurgarn, line 54: Would useCountryValue, in gridstudy-app, of any use (to be transfered to common-ui) ?

src/components/dialogs/generic-filter-dialog.js
```diff
@@ -439,16 +331,37 @@ export const GenericFilterDialog = ({ id, open, onClose, title }) => {
...
+        return value1;
+    }
+
     const editDone = () => {
         let res = {};
-        Object.entries(currentFormEdit).forEach(([key, value]) => {
-            res[key] = value.enabled ? value.value : null;
+        Object.entries(currentFormEdit).forEach(([key, obj]) => {
+            if (
+                key.startsWith('nominalVoltage') && // dont send nominalVoltage with null values
```
laurgarn: The comment seems, at least to my first look, no consistent with the code.
It is written "no null" in the comment while the then block puts null.
Could it be explained more accurately ?

src/components/dialogs/generic-filter-dialog.js
```diff
@@ -473,54 +386,23 @@ export const GenericFilterDialog = ({ id, open, onClose, title }) => {
...
-                                            )}
-                                        </Grid>
-                                    );
-                                })}
-                            </Grid>
-                        );
-                    else return renderFilter(key, definition);
+                    if (definition.occurs) {
+                        return [
+                            ...Array.from(Array(definition.occurs).keys()),
```
laurgarn, line 390: Why is there a need to create a copy of (new) array of indices, just to map those indices (to a new array) ?

src/components/dialogs/generic-filter-dialog.js
```diff
@@ -473,54 +386,23 @@ export const GenericFilterDialog = ({ id, open, onClose, title }) => {
...
+                        ].map((n) => {
+                            return renderFilter(
+                                key + (n + 1).toString(),
```
laurgarn, line 393: I am quite sur thet the toString call is useless.

dbraquart: that was existing code

src/components/dialogs/generic-filter-dialog.js
dbraquart: maybe...that was existing code

src/components/dialogs/filters.js
dbraquart: yes, maybe there is something to do to get a common "selectCountry" component. I'd rather do that in a separate PR.

src/components/dialogs/generic-filter-dialog.js
dbraquart: I did not want to change the Json data model exchanged with the back-end.
In previous code, we could not sent a null object "nominalVoltage" (having props operator, value1, value2)
I removed the toggle, then I have to check myself I dont sent an object having null value1/value2.

EstherDarkish (approved):

src/components/dialogs/filters.js
```diff
@@ -103,9 +109,14 @@ export const RangeSelection = ({ initialValue, onChange, titleMessage }) => {
         setEqualityType(e.target.value);
     }
 
+    const regex = /^[0-9]*[.]?[0-9]*$/;
```
EstherDarkish: In gridstudy, there is this function :
`function isFloatNumber(val) {
    return /^-?[0-9]*[.,]?[0-9]*$/.test(val);
}`
It allows negative numbers and dots/comma instead of only dots.

EstherDarkish (changes requested):

src/components/dialogs/filters.js
EstherDarkish: Moreover, because the keyboard blocks the user's input each stroke if the format is not correct, we can have a bad situation for the user :
Step 1 : the user has "0.56111108" as his current value, but wants "0.50111108" instead.
Step 2 : the user copies " 0.50111108" from some place (with a space before, because that's what users do).
Step 3 : the user pastes it in his field, visually **it seems like** it was pasted, but no, the original value is still there.

dbraquart: A nominal voltage level is a positive value, then the regexp is ok (I added the "," as you noticed).

to enhance the copy/paste issue you mentionned, we display an info snackbar message in case of bad copy.

EstherDarkish (approved):

laurgarn (approved):

PR #154, opened 2022-10-28
fix element draging when trying to select text

YenguiSeddik:
fixed the issue that occurs when trying to select a text (equipment id) in the manual filter table with the mouse, the line is dragged and you cannot select it.

### DIFF
--- a/src/components/dialogs/manual-filter-creation-dialog.js
+++ b/src/components/dialogs/manual-filter-creation-dialog.js
@@ -63,11 +63,9 @@ const ManualFilterRow = ({
         <Draggable draggableId={id} index={index} key={id + value + index}>
             {(provided) => (
                 <div
-                    draggable="true"
-                    ref={provided.innerRef}
                     style={{ width: '100%' }}
+                    ref={provided.innerRef}
                     {...provided.draggableProps}
-                    {...provided.dragHandleProps}
                 >
                     <Grid
                         container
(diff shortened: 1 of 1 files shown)


bendaamerahm (approved):

laurgarn (approved):

src/components/dialogs/manual-filter-creation-dialog.js
```diff
@@ -66,19 +66,18 @@ const ManualFilterRow = ({
...
                         spacing={2}
                         key={index + id + 'container item'}
                         sx={{ width: '100%', height: '50%' }}
-                        draggable="true"
                         ref={provided.innerRef}
-                        {...provided.draggableProps}
-                        {...provided.dragHandleProps}
                     >
-                        <Grid xs={1} item>
+                        <Grid xs={1} item draggable="true">
```
sBouzols: Why to use a direct HTML attribute here ?
You seems to misunderstand the use of Draggable component.
Please check this documentation to apply the correct components prop setting
https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/api/draggable.md

sBouzols (changes requested):

src/components/dialogs/manual-filter-creation-dialog.js
YenguiSeddik: ok. done.

sBouzols (approved):

PR #157, opened 2022-11-02
improve error message when import case file with bad extension

Walid-Sahnoun:
Signed-off-by: walid-sahnoun <walid.sahnoun@rte-france.com>


src/components/dialogs/create-case-dialog.js
```diff
@@ -84,8 +84,14 @@ export function CreateCaseDialog({ onClose, open }) {
             parentDirectoryUuid: activeDirectory,
         })
             .then()
-            .catch((message) => {
-                snackbarMessage(message, 'caseCreationError', { name });
+            .catch(() => {
```
klesaulnier: we are catching all errors here, this REST call could throw an error for other reasons than the file format
Shouldn't we display this error only if we catch the format error ?

src/components/dialogs/create-case-dialog.js
```diff
@@ -84,8 +84,14 @@ export function CreateCaseDialog({ onClose, open }) {
             parentDirectoryUuid: activeDirectory,
         })
             .then()
```
klesaulnier, line 86: Shouldn't we also show this message if we create a study with a file having the wrong format ?

klesaulnier (changes requested):

src/components/dialogs/create-case-dialog.js
bendaamerahm: i agree

src/components/dialogs/create-case-dialog.js
```diff
@@ -85,7 +85,15 @@ export function CreateCaseDialog({ onClose, open }) {
...
             .catch((message) => {
-                snackbarMessage(message, 'caseCreationError', { name });
+                message.trim().startsWith('422')
```
bendaamerahm: i'd prefer to add a constant with code status 422 ` const UNPROCESSABLE_ENTITY_STATUS = '422'`
then unstead of doing this `message.trim().startsWith('422')`  you can do this unstead `message.includes(UNPROCESSABLE_ENTITY_STATUS) `

src/components/dialogs/create-case-dialog.js
bendaamerahm: i think we need to do the same for study for the same error

bendaamerahm (changes requested):

src/components/dialogs/create-case-dialog.js
klesaulnier: You changed backend to throw a 422 http status, check this status instead of parsing the error message

klesaulnier (changes requested):

src/components/dialogs/create-case-dialog.js
Walid-Sahnoun: ok

src/components/dialogs/create-case-dialog.js
Walid-Sahnoun: ok

src/components/dialogs/create-case-dialog.js
Walid-Sahnoun: done

Walid-Sahnoun: done

bendaamerahm (approved):

bendaamerahm (approved):

src/components/dialogs/create-case-dialog.js
```diff
@@ -85,7 +85,15 @@ export function CreateCaseDialog({ onClose, open }) {
         })
             .then()
             .catch((message) => {
-                snackbarMessage(message, 'caseCreationError', { name });
+                message.includes(UNPROCESSABLE_ENTITY_STATUS)
```
klesaulnier: We are still not relying on the status code of the response, which makes the backend modification useless here
Is it possible to fix it ?

klesaulnier (changes requested):

src/components/dialogs/create-case-dialog.js
Walid-Sahnoun: yes

src/utils/rest-api.js
```diff
@@ -432,7 +438,9 @@ export function duplicateCase(
 
     return backendFetch(url, {
         method: 'post',
-    }).then((response) => handleResponse(response, false));
+    })
```
klesaulnier: why are we doing this here ?

src/utils/rest-api.js
```diff
@@ -409,7 +409,13 @@ export function createCase({ name, description, file, parentDirectoryUuid }) {
     return backendFetch(url, {
         method: 'post',
         body: formData,
-    }).then((response) => handleResponse(response, false));
+    }).then((response) =>
+        response.ok
+            ? response
+            : response.status === 422
```
klesaulnier: this is not necessary, you already check status code in create-case-dialog.js, we don't need to do it twice

src/utils/rest-api.js
Walid-Sahnoun: i will fixed

src/utils/rest-api.js
Walid-Sahnoun: ok

klesaulnier (approved):

PR #158, opened 2022-11-04
Add filter creation (automatic and manual) on voltage level and substation

FranckLecuyer:


bendaamerahm (approved):

dbraquart (approved):
all clear.
manual tests OK (creation, update, deletion, export for both auto/manu filters).

PR #164, opened 2022-11-16
reflect websocket changes

AbdelHedhili:


AbdelHedhili:
not needed any longer since we keep the choice of passing parameters through url

PR #165, opened 2022-11-16
Prefill Name field when creating cases or studies

ayolab:

### DIFF
--- a/src/components/dialogs/create-study-dialog.js
+++ b/src/components/dialogs/create-study-dialog.js
@@ -5,13 +5,12 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import makeStyles from '@mui/styles/makeStyles';
 import CheckIcon from '@mui/icons-material/Check';
 import SettingsIcon from '@mui/icons-material/Settings';
 import Button from '@mui/material/Button';
-import TextField from '@mui/material/TextField';
 import Dialog from '@mui/material/Dialog';
 import DialogActions from '@mui/material/DialogActions';
 import DialogContent from '@mui/material/DialogContent';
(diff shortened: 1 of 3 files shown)


src/components/dialogs/create-study-dialog.js
```diff
@@ -325,6 +328,16 @@ export const CreateStudyDialog = ({ open, onClose, providedCase }) => {
         }
     }, [activeDirectory]);
 
+    useEffect(() => {
+        if (studyNameRef.current.trim().length === 0 && selectedFile != null) {
```
etiennehomer: Maybe you can mutualize this code

src/components/dialogs/field-hook.js
```diff
@@ -54,6 +54,10 @@ export const useTextValue = ({
         setValue(event.target.value);
     }, []);
 
+    const handleExternalNameChange = useCallback((value) => {
```
etiennehomer: you can export setValue instead of creating this callback

etiennehomer (changes requested):

etiennehomer:
Maybe in this PR, we can use a fieldHook for studyName field

ayolab:
I reused the nameField hook and the textValue and I created a custom hook for Name field pre-filling

etiennehomer:
Tests OK

src/components/dialogs/field-hook.js
```diff
@@ -250,9 +254,30 @@ export const useNameField = ({
...
+            selectedFile != null
+        ) {
+            setValue(
+                selectedFile.name.substr(0, selectedFile.name.indexOf('.'))
+            );
+        } else if (
+            selectedFile == null &&
+            setValue != null &&
+            setValue !== undefined
+        ) {
```
etiennehomer: just test && setValue

etiennehomer (changes requested):

src/components/dialogs/field-hook.js
etiennehomer: But your test on setValue should be upper nope ?
setValue is also used in the other branch of the if

ayolab: Done !

etiennehomer (approved):

PR #168, opened 2022-11-21
Add creation date column

Meklo:


src/components/directory-content.js
```diff
@@ -577,6 +594,7 @@ const DirectoryContent = () => {
     };
 
     const getCurrentChildrenWithNotClickableRows = () => {
+        console.log(currentChildren);
```
FranckLecuyer: to be removed

FranckLecuyer (changes requested):

src/components/directory-content.js
Meklo: I removed it

Walid-Sahnoun (approved):

FranckLecuyer (approved):

PR #169, opened 2022-11-23
Change instantiation mechanism for pop-up dialogs

ayolab:

### DIFF
--- a/src/components/directory-content.js
+++ b/src/components/directory-content.js
@@ -125,8 +125,7 @@ const DirectoryContent = () => {
     const [mousePosition, setMousePosition] =
         React.useState(initialMousePosition);
 
-    const [dialogToOpen, setDialogToOpen] = useState(null);
-
+    const [openDialog, setOpenDialog] = useState(constants.DialogsId.NONE);
     const handleRowClick = (event) => {
         if (childrenMetadata[event.rowData.elementUuid] !== undefined) {
             const subtype = childrenMetadata[event.rowData.elementUuid].subtype;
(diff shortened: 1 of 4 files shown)


src/components/directory-content.js
```diff
@@ -122,6 +125,8 @@ const DirectoryContent = () => {
     const [mousePosition, setMousePosition] =
         React.useState(initialMousePosition);
 
+    const [dialogToOpen, setDialogToOpen] = useState(null);
```
etiennehomer: Rather [dialogOpen, setDialogOpen] ?

src/components/directory-content.js
```diff
@@ -734,83 +797,48 @@ const DirectoryContent = () => {
...
-                              }
-                            : undefined
-                    }
-                />
+                {openContentMenu && (
+                    <ContentContextualMenu
+                        activeElement={activeElement}
+                        selectedElements={getSelectedChildren()}
+                        open={openContentMenu}
+                        openDialog={openDialog}
```
etiennehomer: It's weird to keep "openContentMenu && " and "open={openContentMenu}".
Maybe for menus, we should keep it working with open={openContentMenu} only as it will be often open and closed

etiennehomer:
Tests OK

src/components/directory-content.js
etiennehomer: @sBouzols what do you think ?

src/components/directory-content.js
ayolab: The state contains the name of the dialog that we want to open, that's why I have chosen "dialogToOpen"

etiennehomer:
There's a warning in the console when we open a directory :
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.

etiennehomer (approved):

sBouzols (approved):
Code Review OK

src/components/directory-content.js
sBouzols: I don't find the `openContentMenu &&` part.
It's ok for me to render both ContentMenu and DirectoryMenu in the DirectoryContent and to control their showing by their open props. It's ok to use controlled framwork for those menu passing openDialog and setOpenDialog as well.
Will see in the future if it's a bad design...

PR #170, opened 2022-11-24
add energy source criteria on group filters

anas-drissi:


src/utils/equipment-types.js
```diff
@@ -13,6 +13,20 @@ const Countries = {
         type: filteredTypes.countries,
     },
 };
+const EnergySource = {
+    energySource: {
+        name: 'EnergySourceText',
+        type: filteredTypes.enum,
+        enumValues: [
+            { id: 'HYDRO', label: 'Hydro' },
```
laurgarn: I would favor the name 'i18n_id' over 'label', as the latter is not a real label, as it as to undergo the internationalisation step.
I would even simplify the mapping aspect, by relying on js object syntax and mecanisms :
```
enumValues : {
  HYDRO : 'Hydro', NUCLEAR: 'Nuclear', // and so on
}
```
that would save you the burden of naming the parts.

src/components/dialogs/filters.js
```diff
@@ -83,6 +83,62 @@ export const CountriesSelection = ({
...
+export const EnumSelection = ({
+    initialValue,
+    onChange,
+    titleMessage,
+    enumValues,
+}) => {
+    const [value, setValue] = useState(initialValue);
+    const intl = useIntl();
+
+    const getEnumTranslation = useCallback(
```
laurgarn: I would split the translation in two phases.
Cache the enumEntry -> localizedLabel in a plain js object,
then use direct it directly in the function passed to the autocomplete.
I would also cache the options given to the autoComplete.
With the simplification proposed in a comment well below this one, it would give smth like :
```
const enumLabels = useMemo(() => Object.fromEntries(
   Object.entries(enumValues)
     .map([k,v] => [k, intl.formatMessage({id:v})]),
   [intl, enumValues]);
const options = useMemo(() => Object.keys(enumValues), [enumValues]);
const getEnumTranslation = useCallback((enumEntry) => enumLabels[enumEntry], [enumLabels] ?? '');
```

laurgarn (approved):
Works

FranckLecuyer (approved):

src/components/dialogs/filters.js
dbraquart: that's a pleasant refactor. I take it.

PR #171, opened 2022-11-25
Last modification date and user columns

klesaulnier:


src/components/directory-content.js
```diff
@@ -342,28 +342,28 @@ const DirectoryContent = () => {
...
             </div>
         );
     }
 
-    function createdCellRender(cellData) {
-        const created = new Date(cellData.rowData[cellData.dataKey]);
-        if (created instanceof Date && !isNaN(created)) {
-            const date = created.toLocaleDateString(intl.locale);
+    function dateCellRender(cellData) {
+        const data = new Date(cellData.rowData[cellData.dataKey]);
```
Meklo, line 366: Calling it data instead of date is deliberate ? What is your train of thought, since you renamed the renderer to `dateCellRender`

niconoir (approved):

src/components/directory-content.js
klesaulnier: there is a date variable two lines below, so I thought it was ok here

Meklo (approved):

Meklo:
When doing a "Replace with script" contextual action from a contingency list, it returns an 400 error with the associated message :
`org.springframework.web.bind.MissingRequestHeaderException: Required request header 'userId' for method parameter type String is not present`

klesaulnier:
> When doing a "Replace with script" contextual action from a contingency list, it returns an 400 error with the associated message : `org.springframework.web.bind.MissingRequestHeaderException: Required request header 'userId' for method parameter type String is not present`

fixed, a change in explore-server was needed

Meklo (approved):

PR #172, opened 2022-11-28
reorganize ihm filtre creation edition

Walid-Sahnoun:


LICENSE
```diff
@@ -232,7 +232,7 @@ recipient of ordinary skill to be able to understand it.
 5. Termination
 --------------
 
-5.1. The rights granted under this License will terminate automatically
+5.1. The rights granted under this License will terminate CriteriaBasedally
```
laurgarn: A search and replace went too far ?

src/components/dialogs/create-filter-dialog.js
```diff
@@ -108,8 +108,33 @@ const CreateFilterDialog = ({
...
-    const [newListType, setNewListType] = useState(FilterType.AUTOMATIC);
-    const [filterType, setFilterType] = useState('');
+    const [newListType, setNewListType] = useState(FilterType.CRITERIA);
+    const [filterType, setFilterType] = useState(null);
+    const [openConfirmationPopup, setopenConfirmationPopup] = useState(false);
+    const [choosedFilterType, setChoosedFilterType] = useState(
+        FilterType.CRITERIA
+    );
+
+    const [filterToSave, setFilterToSave] = useState(null);
```
laurgarn, line 118: Why passing from no blank line between declaration lines to one blank line after every declaration line. I suggest reserving blank lines (one at a time) to separate blocks of more closely related declarations.

src/components/dialogs/create-filter-dialog.js
```diff
@@ -172,15 +197,107 @@ const CreateFilterDialog = ({
...
+        isGeneratorOrLoad,
+        isFilterCreation,
+        equipmentType,
+        name,
+        id
+    ) => {
+        if (
+            tableValues.every((el) => {
+                if (!el?.equipmentID) {
+                    setCreateFilterErr(
```
laurgarn: If we have the situation where we have multiple lines without ID, we set each time same message, telling nothing about which line(s) is/are ill-formed. Sadly.

src/components/dialogs/create-filter-dialog.js
```diff
@@ -108,8 +108,33 @@ const CreateFilterDialog = ({
...
+        FilterType.CRITERIA
+    );
+
+    const [filterToSave, setFilterToSave] = useState(null);
+
+    const [tableValues, setTableValues] = useState([]);
+
+    const [isGeneratorOrLoad, setIsGeneratorOrLoad] = useState(false);
+
+    const [isFilterCreation, setIsFilterCreation] = useState(false);
```
laurgarn: As the context is already about filtering, you could just go with isCreation.

src/components/dialogs/create-filter-dialog.js
```diff
@@ -108,8 +108,33 @@ const CreateFilterDialog = ({
...
+    const [openConfirmationPopup, setopenConfirmationPopup] = useState(false);
+    const [choosedFilterType, setChoosedFilterType] = useState(
+        FilterType.CRITERIA
+    );
+
+    const [filterToSave, setFilterToSave] = useState(null);
+
+    const [tableValues, setTableValues] = useState([]);
+
+    const [isGeneratorOrLoad, setIsGeneratorOrLoad] = useState(false);
```
laurgarn, line 120: The 'or' can be interpreted as a real 'or', or it can be interpreted as a mean to explicit the opposition .
'Over' could be used in the latter case, but here, we would get 'isGeneratorOverLoad' where one could miss the lower 'l' and read 'Overload' that means something different ! So 'isForGenerator' would do shorter and not worse.
For the former interpretation, isInjection would be better, though it means redundancy (and possibly inconsistency) with equipmentType.

src/components/dialogs/create-filter-dialog.js
```diff
@@ -172,15 +197,107 @@ const CreateFilterDialog = ({
 
     const resetDialog = () => {
         setNewListName('');
-        setNewListType(FilterType.AUTOMATIC);
+        setNewListType(FilterType.CRITERIA);
         setFilterType('');
         setLoadingCheckFilterName(false);
         setCreateFilterErr('');
         setFilterNameValid(false);
+        setopenConfirmationPopup(false);
```
laurgarn: The lower 'o' of ''open' in 'setopenConfirmationPopup' should have been an upper 'O'

src/components/dialogs/create-filter-dialog.js
```diff
@@ -172,15 +197,107 @@ const CreateFilterDialog = ({
...
+        ) {
+            if (isGeneratorOrLoad) {
+                // we check if all the distribution keys are null.
+                // If one is set, all the distribution keys that are null take 0 as value
+                let isAllKeysNull = tableValues.every(
+                    (row) => !row.distributionKey
+                );
+                tableValues.forEach((val, index) => {
+                    if (!isAllKeysNull && !val.distributionKey) {
+                        tableValues[index] = {
```
laurgarn: `tableValues[index][distributionKey] = 0;`
could do, no ?

src/components/dialogs/create-filter-dialog.js
```diff
@@ -108,8 +108,33 @@ const CreateFilterDialog = ({
     const classes = useStyles();
     const intl = useIntl();
     const timer = React.useRef();
-    const [newListType, setNewListType] = useState(FilterType.AUTOMATIC);
-    const [filterType, setFilterType] = useState('');
+    const [newListType, setNewListType] = useState(FilterType.CRITERIA);
+    const [filterType, setFilterType] = useState(null);
+    const [openConfirmationPopup, setopenConfirmationPopup] = useState(false);
```
laurgarn: Why don't you stick with the convention of prefixing 'is' for boolean state, here ?
I suggest isConfirmationPopupOpen + setConfirmationPopupOpen.

src/components/dialogs/create-filter-dialog.js
```diff
@@ -265,26 +438,39 @@ const CreateFilterDialog = ({
                         aria-label="type"
                         name="filterType"
                         value={newListType}
-                        defaultValue={FilterType.AUTOMATIC}
-                        onChange={(e) => setNewListType(e.target.value)}
+                        defaultValue={FilterType.CRITERIA}
+                        onChange={(e) => handleRadioButtonClick(e)}
```
laurgarn: 'handleRadioButtonClick' -> 'onFilterTypeChange'

src/translations/fr.json
```diff
@@ -178,5 +178,8 @@
     "uploadingElement": "Téléversement en cours : pas d'opération autorisée",
     "serverConnectionFailed" : "Echec de connexion avec le serveur. Veuillez réessayer ultérieurement",
     "invalidFormatOrName":"Format ou nom du fichier importé non valide",
-    "parameterLoadingProblem":"Probléme de chargement des paramétres"
+    "parameterLoadingProblem":"Probléme de chargement des paramétres",
```
dbraquart: Probléme -> Problème
paramétres -> paramètres

src/translations/fr.json
```diff
@@ -178,5 +178,8 @@
...
+    "dragAndDrop": "glisser déposer",
+    "changeTypeMessage": "Voulez-vous change le type ?",
```
dbraquart: change -> changer

src/components/dialogs/create-filter-dialog.js
```diff
@@ -235,6 +365,49 @@ const CreateFilterDialog = ({
...
+        setChoosedFilterType(event.target.value);
+    };
+
+    const handlePopupConfirmation = () => {
+        setopenConfirmationPopup(false);
+        setNewListType(choosedFilterType);
+        setEquipmentType(null);
+    };
+
+    const renderchangeFilterTypePopup = () => {
```
laurgarn: renderchangeFilterTypePopup -> renderChangeFilterTypePopup

src/translations/fr.json
```diff
@@ -178,5 +178,8 @@
     "uploadingElement": "Téléversement en cours : pas d'opération autorisée",
     "serverConnectionFailed" : "Echec de connexion avec le serveur. Veuillez réessayer ultérieurement",
```
laurgarn: Echec -> Échec

src/translations/fr.json
```diff
@@ -178,5 +178,8 @@
...
     "invalidFormatOrName":"Format ou nom du fichier importé non valide",
-    "parameterLoadingProblem":"Probléme de chargement des paramétres"
+    "parameterLoadingProblem":"Probléme de chargement des paramétres",
+    "dragAndDrop": "glisser déposer",
+    "changeTypeMessage": "Voulez-vous change le type ?",
+    "keepCSVDataMessage": "Voulez-vous ajouter les critères ou remplacer ceux existants ?"
```
laurgarn: -> to be displayed with three buttons : "cancel", "add", "replace".

src/components/dialogs/explicit-naming-filter-creation-dialog.js
```diff
@@ -0,0 +1,196 @@
...
+    };
+
+    useEffect(() => {
+        setCreateFilterErr('');
+    }, [tableValues]);
+
+    const handleCreateFilter = () => {
+        if (
+            tableValues.every((el) => {
+                if (!el?.equipmentID) {
```
YenguiSeddik: why not directly use if(tableValues.every((el) => !el?.equipmentID)) ? this will check every element has equipment id

src/translations/fr.json
Walid-Sahnoun: done

src/translations/fr.json
Walid-Sahnoun: done

LICENSE
Walid-Sahnoun: yes i fix it

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: done

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: well i think we didn't need to know about which line is ill-formed.

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: ok

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: here we need to know the equipment is generator or load

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: ok

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: here i need to assign the hole object

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: done

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: done

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: done

src/translations/fr.json
Walid-Sahnoun: done

src/components/dialogs/create-filter-dialog.js
laurgarn: The whole object ? Why ?

src/translations/fr.json
Walid-Sahnoun: ok i change the dialog with 3 buttons

src/components/dialogs/explicit-naming-filter-creation-dialog.js
Walid-Sahnoun: here i need to check every id else i display error message

laurgarn: I don't like this mixing of 'every' with just 'last'.
setCreateFilterErr is a side-effect.
It should not be call in the expressions of the 'if'.
Please call this side effect in its own branch of the if.
A way to do so is
```
const hasMissingId = tableValues.some((el) => !el?.equipmentID);
if (hasMissingId) {
  setCreateFilterErr(
      intl.formatMessage({
          id: 'missingEquipmentsIdsError',
      })
  );
} else {
  // your previous 'then'
}
```
Of course you can inline hasMissingId, if you prefer.

src/translations/fr.json
```diff
@@ -176,7 +176,13 @@
...
     "notElementCreator": "Vous n'êtes pas le créateur de l'élément",
     "uploadingElement": "Téléversement en cours : pas d'opération autorisée",
-    "serverConnectionFailed" : "Echec de connexion avec le serveur. Veuillez réessayer ultérieurement",
+    "serverConnectionFailed" : "Échec de connexion avec le serveur. Veuillez réessayer ultérieurement",
     "invalidFormatOrName":"Format ou nom du fichier importé non valide",
-    "parameterLoadingProblem":"Probléme de chargement des paramétres"
+    "parameterLoadingProblem":"Problème de chargement des paramètres",
+    "dragAndDrop": "glisser déposer",
+    "changeTypeMessage": "Voulez-vous changer le type ?",
+    "keepCSVDataMessage": "Voulez-vous  remplacer ou ajouter vos identifiants ?",
```
dbraquart: double space between vous/remplacer

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: well i was wrong i can use this tableValues[index]['distributionKey'] = 0;

src/components/dialogs/explicit-naming-filter-creation-dialog.js
Walid-Sahnoun: ok i changed like this

src/components/dialogs/criteria-based-filter-dialog.js
```diff
@@ -0,0 +1,117 @@
...
+import React, { useEffect, useRef, useState } from 'react';
+import makeStyles from '@mui/styles/makeStyles';
+import DialogTitle from '@mui/material/DialogTitle';
+import DialogContent from '@mui/material/DialogContent';
+import DialogActions from '@mui/material/DialogActions';
+import Button from '@mui/material/Button';
+import Dialog from '@mui/material/Dialog';
+import { saveFilter, saveFormContingencyList } from '../../utils/rest-api';
+import { ElementType } from '../../utils/elementType';
+import {} from '../../utils/equipment-types';
```
dbraquart: surprising. What import is it ?

src/components/dialogs/field-hook.js
```diff
@@ -333,13 +354,38 @@ export const useEquipmentTableValues = ({
...
     const handleChangeOrder = useCallback(
-        (index, direction) => {
+        (direction) => {
             const res = [...values];
-            const [item] = res.splice(index, 1);
-            res.splice(index + direction, 0, item);
+            const result = [...selectedIds];
+            result.sort();
+
+            let isContigue = true;
```
dbraquart: isContiguous in english

src/components/dialogs/field-hook.js
```diff
@@ -419,19 +534,93 @@ export const useEquipmentTableValues = ({
...
+                            <ArrowCircleUp />
+                        </IconButton>
+                        <IconButton
+                            key={id + name + 'downButton'}
+                            disabled={
+                                selectedIds.size === 0 ||
+                                selectedIds.has(values.length - 1)
+                            }
+                            onClick={() => {
+                                handleChangeOrder(id, 1);
```
dbraquart: here we passe 2 args to handleChangeOrder(), "down" and 1
While it accepts only one.
Strange

dbraquart (changes requested):
Code review: a couple of remarks to check. Could require a change.

Tests:
Works well, but I found one error case:
- when we create a new automatic/Criteria Filter with 'Lines' type : nominalVoltage1 is not sent in the POST request = > lost.
(same for countries1 and countries2)

A question:
We can't open an existing Manual/Naming Filter anymore (created with a baseline before this current PR). Is it normal ? If yes, dont forget to warn people.

Warning browser console:
- when we create a Filter and  change type (the very first time?):
	16:58:21,644 Warning: Failed prop type: The prop `spacing` of `Grid` can only be used together with the `container` prop.

src/components/dialogs/field-hook.js
Walid-Sahnoun: ok

src/components/dialogs/field-hook.js
Walid-Sahnoun: ok i fix it to one argument

src/components/dialogs/create-filter-dialog.js
```diff
@@ -108,8 +108,29 @@ const CreateFilterDialog = ({
...
+    const [equipmentType, setEquipmentType] = useState(null);
+    const [name, setName] = useState('');
+    const [id, setId] = useState('');
+    const handleCallback = (criteriaFilter) => {
+        if (criteriaFilter) {
+            setFilterToSave(criteriaFilter);
+        }
+    };
+
+    const handleEquipementChange = (equipmentType) => {
```
dbraquart: handleEquipementChange -> handleEquipmentChange
please search everywhere (several occurrences)

dbraquart (changes requested):
test KO: update a contingency list or filter:
when we change the eqpt type, the combo box is not closed (a second click is required). And we have a console warning:
	Uncaught TypeError: handleEquipementChange is not a function
		changeFilterType criteria-filter-dialog-content.js:201

src/components/dialogs/create-filter-dialog.js
Walid-Sahnoun: done

dbraquart (approved):

src/components/dialogs/field-hook.js
```diff
@@ -308,56 +323,163 @@ export const useEquipmentTableValues = ({
...
+        if (selectedIds.size === 0) {
+            selectAllItems();
+        } else if (selectedIds.size === values.length) {
+            setSelectedIds(new Set());
+        } else {
+            selectAllItems();
+        }
+
+        function selectAllItems() {
+            let array = [];
```
laurgarn: `const array = new Set(values.map((v, i) => i));`
;-)

src/components/dialogs/field-hook.js
```diff
@@ -308,56 +323,163 @@ export const useEquipmentTableValues = ({
...
+                let objects = Object.keys(csvData).map(function (key) {
+                    return {
+                        equipmentID: csvData[key][0].trim(),
+                        distributionKey:
+                            csvData[key][1].trim() !== ''
+                                ? csvData[key][1]
+                                : undefined,
+                    };
+                });
+                objects.forEach((elem) => {
```
laurgarn: `values.push(...objects);`

src/components/dialogs/field-hook.js
```diff
@@ -308,56 +323,163 @@ export const useEquipmentTableValues = ({
...
+            }
+            setSelectedIds(new Set(array));
+        }
+    }, [selectedIds.size, values.length]);
+
+    const updateTableValues = useCallback(
+        (csvData, keepTableValues) => {
+            if (csvData) {
+                if (!keepTableValues) {
+                    while (values.length !== 0) {
```
laurgarn: `values.splice(0);`

src/components/dialogs/field-hook.js
```diff
@@ -423,19 +558,93 @@ export const useEquipmentTableValues = ({
...
+                            className={classes.iconColor}
+                            onClick={() => handleDeleteItem()}
+                            disabled={selectedIds.size === 0}
+                        >
+                            <DeleteIcon />
+                        </IconButton>
+                        <IconButton
+                            key={id + name + 'upButton'}
+                            disabled={
+                                selectedIds.size === 0 || selectedIds?.has(0)
```
laurgarn: I would go with :
`!selectedIds?.size || selectedIds.has(0)`

laurgarn (approved):

src/components/dialogs/field-hook.js
```diff
@@ -308,56 +323,163 @@ export const useEquipmentTableValues = ({
...
+                selectedIds.forEach((elem) => {
+                    const [item] = res.splice(elem, 1);
+                    res.splice(elem + direction, 0, item);
+                });
+            }
+            let array = [...selectedIds];
+            array.sort();
+            for (let i = 0; i < array.length; i++) {
+                array[i] = array[i] + direction;
+            }
```
YenguiSeddik: this can be like this: 
const array = Array.from(selectedIds)
                .sort()
                .map((val) => val + direction);

src/components/dialogs/field-hook.js
```diff
@@ -308,56 +323,163 @@ export const useEquipmentTableValues = ({
...
             );
             setValues(res);
+            let array = [];
+            let j = 0;
+            for (let i = 0; i < res.length; i++) {
+                if (res[i].isChecked) {
+                    array[j] = i;
+                    j++;
+                }
+            }
```
YenguiSeddik: Maybe use map to get array from res:
const array = res.filter((val) => val.isChecked)
                .map((val) => {
                return res.indexOf(val);
            })

src/components/dialogs/field-hook.js
Walid-Sahnoun: ok

src/components/dialogs/field-hook.js
Walid-Sahnoun: done

src/components/dialogs/field-hook.js
Walid-Sahnoun: done

src/components/dialogs/field-hook.js
Walid-Sahnoun: done

src/components/dialogs/field-hook.js
Walid-Sahnoun: done

src/components/dialogs/field-hook.js
Walid-Sahnoun: done

YenguiSeddik (approved):

PR #173, opened 2022-11-29
Unify exception handling

ayolab:


src/components/tree-views-container.js
```diff
@@ -473,9 +473,12 @@ const TreeViewsContainer = () => {
                     // Update Tree Map data
                     updateMapData(nodeId, childrenToBeInserted);
                 })
-                .catch((reason) => {
+                .catch((error) => {
                     console.warn(
-                        "Could not update subs of '" + nodeId + "' :" + reason
+                        "Could not update subs of '" +
```
laurgarn: Could use string interpolation ticks : 
```
`Could not update subs of '${nodeId}' : ${error.message}`
```

src/utils/rest-api.js
```diff
@@ -293,30 +315,19 @@ function getElementsIdsListsQueryParams(ids) {
...
     const fetchElementsInfosUrl =
         PREFIX_EXPLORE_SERVER_QUERIES +
         '/v1/explore/elements/metadata' +
         getElementsIdsListsQueryParams(ids);
-    return backendFetch(fetchElementsInfosUrl, {
-        method: 'GET',
-    }).then((response) => handleJsonResponse(response));
+    return backendFetch(
+        fetchElementsInfosUrl,
+        {
```
laurgarn: We may get rid of this method overriding that bring nothing as it is the default.

src/utils/rest-api.js
```diff
@@ -54,16 +54,35 @@ export function connectNotificationsWsUpdateConfig() {
...
+            return Promise.reject({
+                message: text ? text : response.statusText,
+                status: response.status,
+                statusText: response.statusText,
+            });
+        });
+    }
+}
+
+function backendFetch(url, init, isJson, withAuth = true) {
```
laurgarn: I suggest inverting order between 'init' and 'isJson' (that I would call 'expectsJson', did I already tell it ? ;-) ) as the shorter boolean and yet mandatory literal would provoke less reindenting of the numerous calls downward in the diff. 'init' could then even be declared with the _default_ value of undefined bringing less noise in the calls.

SlimaneAmar (approved):

laurgarn (approved):

jonenst (approved):

PR #191, opened 2023-01-10
fix many snackbar when importing studies which fails

Walid-Sahnoun:
Signed-off-by: walid-sahnoun <walid.sahnoun@rte-france.com>


laurgarn (approved):
Test OK

etiennehomer (approved):

PR #192, opened 2023-01-10
Fix Authentification: check if user is null OR user token is expired to call signinSilent

sBouzols:
Signed-off-by: sBouzols <sylvain.bouzols@gmail.com>


sBouzols:
waiting for the newt version of commons-ui

jonenst (approved):

PR #196, opened 2023-01-13
[WIP] Fix create study from DIE with all extensions

YenguiSeddik:
For the moment, we will add only the supported extensions when creating a study from Arcade. When user select "all" in the extension parameter (in Import parameters) we send all the supported extensions.


laurgarn (approved):
Test OK

geofjamg (changes requested):
Waiting for further discussions about how to implement this feature in a cleanest way (back and not open source)

PR #197, opened 2023-01-18
Adapt usePrefillNameField hook to take cases into account

Meklo:


src/components/dialogs/create-study-dialog.js
```diff
@@ -172,7 +172,9 @@ export const CreateStudyDialog = ({ open, onClose, providedCase }) => {
     const [isUploadingFileInProgress, setUploadingFileInProgress] =
         useState(false);
 
-    const [fileCheckedCase, setFileCheckedCase] = useState(false);
+    const [fileCheckedCase, setFileCheckedCase] = useState(
+        providedCase ? true : false
```
anistouri: `providedCase` is always undefined because it's not passed with the parameters when this component is called `<CreateStudyForm open={true} onClose={handleCloseDialog} />`

Meklo: Are you sure ? On main CreateStudyDialog is instanciated this way in `content-contextual-menu.js` :
`<CreateStudyDialog
  open={true}
  onClose={handleCloseDialog}
  providedCase={activeElement}
/>`

anistouri: you're right. You can replace this line by const [fileCheckedCase, setFileCheckedCase] = useState(!!providedCase);

Meklo: Agreed, I did implement it

anistouri (approved):

src/components/dialogs/field-hook.js
```diff
@@ -292,23 +292,31 @@ export const usePrefillNameField = ({
...
     selectedFileOk,
-    createStudyErr,
+    creationError,
     fileCheckedCase,
     touched,
 }) => {
     useEffect(() => {
         if (setValue) {
+            //here selectedFile is a file the user choosed through a picker
             if (
```
etiennehomer, line 302: If the name was prefilled but not touched by the user, we could change it after we change the case file nope ?

Meklo: It's now implemented

etiennehomer (approved):

PR #200, opened 2023-01-24
Fix unmount of CreateStudyDialog

flomillot:
Since we unmount dialogs instead of hiding them, some useEffect didn't work anymore.
Fix : delete the case created for a study if we cancel its creation + double creation of the case when we open the dialog again.
Remove the useEffect of useFileValue and use a callback instead

Signed-off-by: Florent MILLOT <millotflo@gmail.com>

### DIFF
--- a/src/components/dialogs/create-case-dialog.js
+++ b/src/components/dialogs/create-case-dialog.js
@@ -5,7 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useEffect, useRef } from 'react';
 import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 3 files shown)


src/components/dialogs/create-case-dialog.js
```diff
@@ -40,14 +40,12 @@ export function CreateCaseDialog({ onClose, open }) {
...
     const userId = useSelector((state) => state.user.profile.sub);
     const dispatch = useDispatch();
-    const [triggerReset, setTriggerReset] = useState(true);
     const [name, NameField, nameError, nameOk, setCaseName, touched] =
         useNameField({
             label: 'nameProperty',
             autoFocus: true,
             elementType: ElementType.CASE,
             parentDirectoryId: activeDirectory,
-            triggerReset,
```
jonenst, line 50: but then this field is no longer reset ?

flomillot: It was never reset except when the dialog was closed, and now it's unmounted

flomillot: I don't understand why we reset by changing a variable instead of by a callback btw, is there a reason ?

jonenst (approved):

PR #210, opened 2023-02-14
Adjust columns width

Walid-Sahnoun:


src/components/directory-content.js
```diff
@@ -648,50 +648,49 @@ const DirectoryContent = () => {
                             dataKey: 'selected',
                             label: '',
                             headerRenderer: selectionHeaderRenderer,
-                            maxWidth: 42,
+                            minWidth: '4ex',
```
FranckLecuyer, line 651: just a question : why using 'ex' here for width property, because 'ex' is relative to the x-height of the current font

FranckLecuyer (approved):
Tests: OK

ayolab (approved):

src/components/directory-content.js
Walid-Sahnoun: well i use 'ex' to specify the width based on number of characters
example: '20ex'  that's mean allow me a space can contain 20 characters

PR #213, opened 2023-03-08
Create a new component specific to the script dialog's content.

EstherDarkish:
Create a new component to distinguish the script dialog from its content. The goal is to re-use the content in another form.


src/components/dialogs/script-dialog-content.js
```diff
@@ -0,0 +1,110 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
```
sBouzols: 2023

sBouzols (approved):

src/components/dialogs/script-dialog-content.js
EstherDarkish: Thanks, fixed.

PR #218, opened 2023-03-13
[DO_NOT_MERGE] WIP Fixes filters' popup confirmation Enter keypress

EstherDarkish:
Fixes the inconsistent behavior when pressing the Enter key with the filter's confirmation popups. Now, the behavior is the same as clicking on the Validate button.


EstherDarkish:
This fix is not needed anymore, another fix solves this problem by removing the related "Enter key" code.

PR #219, opened 2023-03-14
Removes the deprecated onKeyPress code

EstherDarkish:
Removes the deprecated onKeyPress code that forced popup validations with the Enter key


src/components/dialogs/csv-import-filter-creation-dialog.js
```diff
@@ -131,7 +131,6 @@ const CsvImportFilterCreationDialog = ({
                 <Dialog
                     open={isConfirmationPopupOpen}
                     aria-labelledby="dialog-confirmation-csv-data"
-                    onKeyPress={() => handleAddPopupConfirmation()}
```
flomillot, line 134: Maybe delete the associated function ?

```
    const handleAddPopupConfirmation = () => {
        handleCreateFilter(true);
        setOpenConfirmationPopup(false);
    };
```

EstherDarkish: No, it is used on line 153 : 
`onClick={() => handleAddPopupConfirmation()}`

sBouzols (approved):
Code Review OK
Tests OK
Console Warning checks OK

src/components/dialogs/csv-import-filter-creation-dialog.js
flomillot: Ah yes sorry

flomillot (approved):

PR #223, opened 2023-04-03
Contingency list bugfix

EstherDarkish:
- Stops duplicates and empty values in contingency list equipments.
- Fixes the test that check if there are values in the table before displaying the confirmation popup to replace or add new data.
- Adds better comments in the CSV file for contingency lists.

### DIFF
--- a/src/components/dialogs/csv-import-filter-creation-dialog.js
+++ b/src/components/dialogs/csv-import-filter-creation-dialog.js
@@ -59,13 +59,14 @@ const CsvImportFilterCreationDialog = ({
 
     const csvData = () => {
         let newData = [...data];
-        // Adds 10 empty lines in the CSV file before adding the comment.
-        for (let i = 0; i < 10; i++) {
-            newData.push([]);
+        if (formType === ElementType.CONTINGENCY_LIST) {
+            newData.push(
+                [intl.formatMessage({ id: 'CSVFileCommentContingencyList1' })],
+                [intl.formatMessage({ id: 'CSVFileCommentContingencyList2' })],
+                [intl.formatMessage({ id: 'CSVFileCommentContingencyList3' })],
+                [intl.formatMessage({ id: 'CSVFileCommentContingencyList4' })]
+            );
         }
-        newData.push([
-            intl.formatMessage({ id: 'CSVFileCommentContingencyList' }),
-        ]);
         return newData;
     };
 
@@ -212,6 +213,7 @@ const CsvImportFilterCreationDialog = ({
                                                 ? 'filterCreation'
                                                 : 'contingencyListCreation'
                                         }
+                                        separator={','}
                                     >
                                         <Button variant={'contained'}>
                                             <FormattedMessage id="GenerateCSV" />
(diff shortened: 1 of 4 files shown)


src/components/dialogs/csv-import-filter-creation-dialog.js
```diff
@@ -59,13 +59,14 @@ const CsvImportFilterCreationDialog = ({
 
     const csvData = () => {
         let newData = [...data];
-        // Adds 10 empty lines in the CSV file before adding the comment.
-        for (let i = 0; i < 10; i++) {
-            newData.push([]);
+        if (formType === ElementType.CONTINGENCY_LIST) {
```
etiennehomer, line 62: You don't want to keep some empty rows before the comments ?
It's easier for the user not having to insert new rows

src/translations/en.json
```diff
@@ -118,7 +118,10 @@
     "ImportCSV": "Import CSV",
     "CsvFileName": "CSV file name",
     "CSVFileComment": "# Supported Equipments Types : ",
-    "CSVFileCommentContingencyList": "# Use \"|\" as a separator if a contingency contains several equipments",
+    "CSVFileCommentContingencyList1": "# Use \"|\" as a separator if a contingency contains several equipments",
+    "CSVFileCommentContingencyList2": "# Line 1,ID_line1",
```
etiennehomer, line 122: Maybe you can remove "# Transformer 2,ID_transformer2"
And replace "# Line 1,ID_line1" by "# N-1 on line, ID_line1"

src/components/dialogs/csv-import-filter-creation-dialog.js
```diff
@@ -127,14 +128,23 @@ const CsvImportFilterCreationDialog = ({
     };
 
     const handleOpenCSVConfirmationDataDialog = () => {
-        const res = tableValues.filter(
-            (d) => !Object.values(d).every((v) => v == null)
-        ).length;
-        if (res === 0) {
+        // We check if there are values in the table
+        const isValuesInTable = tableValues.some(
```
etiennehomer, line 132: rename for isTableFIlled ? areSomeValuesPresent ?

src/components/dialogs/csv-import-filter-creation-dialog.js
EstherDarkish: Why can't the user insert his data bellow the comments ? Anyway, I'll see with the product owner.

src/translations/en.json
EstherDarkish: These comments are from the Product Owner

src/components/dialogs/csv-import-filter-creation-dialog.js
EstherDarkish: isTableFilled won't work because it implies the table has no empty cell (which this code do not test), and I'd rather keep the "isXXX" convention for boolean values.

etiennehomer (approved):

src/translations/en.json
etiennehomer: ok

YenguiSeddik (approved):

PR #227, opened 2023-04-17
Adds the esLint rule to force curly braces in statements.

EstherDarkish:
Added a new esLint rule to force curly braces in statements.

To fix and update the code, I used Intellij's `Code => Analyse Code => Run Inspection by Name => "Statement body without braces" `on Gridexplore's *.js files to find these violations.

### DIFF
--- a/src/components/dialogs/create-filter-dialog.js
+++ b/src/components/dialogs/create-filter-dialog.js
@@ -122,10 +122,8 @@ const CreateFilterDialog = ({
     const handleCallback = (criteriaFilter, veto) => {
         if (veto) {
             setFilterToSave(null);
-        } else {
-            if (criteriaFilter) {
-                setFilterToSave(criteriaFilter);
-            }
+        } else if (criteriaFilter) {
+            setFilterToSave(criteriaFilter);
         }
     };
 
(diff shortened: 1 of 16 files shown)


src/components/dialogs/create-filter-dialog.js
```diff
@@ -120,9 +120,12 @@ const CreateFilterDialog = ({
     const [name, setName] = useState('');
     const [id, setId] = useState('');
     const handleCallback = (criteriaFilter, veto) => {
-        if (veto) setFilterToSave(null);
-        else if (criteriaFilter) {
-            setFilterToSave(criteriaFilter);
+        if (veto) {
+            setFilterToSave(null);
+        } else {
```
laurgarn: `if else` stair

laurgarn (approved):

src/components/dialogs/create-filter-dialog.js
etiennehomer: keep the 'else if' nope ?
it's a warning for eslint ?

etiennehomer (approved):

src/components/dialogs/create-filter-dialog.js
EstherDarkish: Fixed. Intellij was overzealous

PR #232, opened 2023-05-03
make all element types directory hovered

ghazwarhili:


FranckLecuyer (approved):
Code review: OK
Test: OK

Walid-Sahnoun (approved):

src/components/directory-content.js
```diff
@@ -621,7 +621,7 @@ const DirectoryContent = () => {
         return currentChildren.map((child) => {
             return {
                 ...child,
-                notClickable: child.type === ElementType.CASE,
+                notClickable: false,
```
sBouzols, line 624: I don't think it's a good idea to set notClickable to false for every children.
What the point of this function getCurrentChildrenWithNotClickableRows otherwise ?
Clicking on a case doesn't do anything, and we don't want to show a pointing cursor on this line for example.
I think the correct fix should be in commons-ui virtualized table to apply `tableRowHover` CSS class to the row even if it's not clickable.

sBouzols (changes requested):
This seems not the a good fix

PR #234, opened 2023-05-11
Debounce filter & contingency creation, useNameField

ghazwarhili:


src/components/dialogs/create-contingency-list-dialog.js
```diff
@@ -98,58 +98,60 @@ export const CreateContingencyListDialog = ({ open, onClose }) => {
...
-        } else {
-            setContingencyFormState('', false);
-            setLoadingCheckContingencyName(false);
-        }
-    };
+        },
+        [activeDirectory, intl]
+    );
 
+    const debouncedUpdateContingencyFormState = useMemo(
```
sBouzols: useCallback

src/components/dialogs/create-filter-dialog.js
```diff
@@ -136,54 +135,57 @@ const CreateFilterDialog = ({
...
-        } else {
-            setFilterFormState('', false);
-            setLoadingCheckFilterName(false);
-        }
-    };
+        },
+        [activeDirectory, intl]
+    );
+
+    const debouncedUpdateFilterFormState = useMemo(
```
sBouzols: useCallback

src/components/dialogs/field-hook.js
```diff
@@ -237,6 +231,11 @@ export const useNameField = ({
         ]
     );
 
+    const debouncedUpdateValidity = useMemo(
```
sBouzols: useCallback

sBouzols (changes requested):

src/components/dialogs/create-contingency-list-dialog.js
sBouzols: nope useMemo is better you are right

src/components/dialogs/create-filter-dialog.js
sBouzols: nope useMemo is better you are right

src/components/dialogs/field-hook.js
sBouzols: nope useMemo is better you are right

sBouzols:
Move your custom hook in commons-ui then use it here and all will be good :)
Same naming convention could be nice : `debounced...`

jonenst (approved):

sBouzols (approved):

PR #243, opened 2023-06-21
Fix: empty all fields after case creation.

AAJELLAL:


Walid-Sahnoun (approved):
review ok
tests ok

src/components/dialogs/field-hook.js
```diff
@@ -137,6 +141,9 @@ export const useFileValue = ({ fileExceedsLimitMessage, isLoading }) => {
             setFileOk(false);
         }
     }, [selectedFile, fileExceedsLimitMessage, intl]);
+    if (noSetFileNeeded) {
```
SlimaneAmar: Don't understand

SlimaneAmar (changes requested):

src/components/dialogs/field-hook.js
AAJELLAL: I used it in order to return specific fields. But as discussed there is another way to do it.

flomillot:
Sorry but I'm pretty busy

SlimaneAmar (approved):

SlimaneAmar:
review ok
tests ok

Walid-Sahnoun (approved):
code ok 
test ok

PR #250, opened 2023-07-19
Fix keyboard key press

YenguiSeddik:
Fix Keyboard buttons behavior (arrow left, arrow right) when are used inside "Equipments" column in explicit naming contingency list

### DIFF
--- a/src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js
+++ b/src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js
@@ -28,11 +28,13 @@ export const getExplicitNamingSchema = (id) => ({
         })
     ),
 });
-const suppressEnter = (params) => {
+const suppressKeyboardKey = (params) => {
     const KEY_ENTER = 'Enter';
+    const KEY_ARROW_LEFT = 'ArrowLeft';
+    const KEY_ARROW_RIGHT = 'ArrowRight';
     const event = params.event;
     const key = event.key;
-    return key === KEY_ENTER;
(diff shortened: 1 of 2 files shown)


src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js
```diff
@@ -28,11 +28,15 @@ export const getExplicitNamingSchema = (id) => ({
         })
     ),
 });
-const suppressEnter = (params) => {
+const suppressKeyboardEvent = (params) => {
     const KEY_ENTER = 'Enter';
```
flomillot: You create 5 const but honestly imo it's useless here

```suggestion
    const key = params.event.key
    return key === 'Enter' || key === 'ArrowLeft' || key === 'ArrowRight'
```

Tristan-WorkGH: it's how it is shown in [the documentation](https://www.ag-grid.com/react-data-grid/keyboard-navigation/#suppress-keyboard-events), to use a `const KEY_*`

Tristan-WorkGH (approved):
good for me

src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js
flomillot: It's because they are reused sometimes. Here I don't see the point. But not a big deal anyway.

flomillot (approved):

PR #251, opened 2023-07-21
Fix contingency list confirmation popup

YenguiSeddik:

PR #252, opened 2023-07-21
Refactor filter dialogs to use react-hook-form and AGGrid

flomillot:


src/components/dialogs/commons/criteria-based/criteria-based-utils.js
```diff
@@ -106,26 +135,79 @@ export const CONTINGENCY_LIST_EQUIPMENTS = {
...
-        ...getRangeInputDataForm(NOMINAL_VOLTAGE_1, nominalVoltage1),
-        ...getRangeInputDataForm(NOMINAL_VOLTAGE_2, nominalVoltage2),
+export const getCriteriaBasedFormData = (criteriaValues, extraFields) => ({
+    [CRITERIA_BASED]: {
+        [COUNTRIES]: criteriaValues?.[COUNTRIES] ?? [],
+        [COUNTRIES_1]: criteriaValues?.[COUNTRIES_1] ?? [],
+        [COUNTRIES_2]: criteriaValues?.[COUNTRIES_2] ?? [],
+        ...getRangeInputDataForm(
+            NOMINAL_VOLTAGE,
+            criteriaValues?.[NOMINAL_VOLTAGE] ?? DEFAULT_RANGE_VALUE
```
klesaulnier:
```suggestion
            criteriaValues?.[NOMINAL_VOLTAGE]
```
DEFAULT_RANGE_VALUE is already the default value
Same for calls below

src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js
```diff
@@ -108,6 +108,7 @@ const ExplicitNamingForm = () => {
             name={EQUIPMENT_TABLE}
             columnDefs={columnDefs}
             defaultRowData={DEFAULT_ROW_VALUE}
+            defaultEmptyRowsNumber={3}
```
klesaulnier: I think we should handle this with RHF only

src/components/dialogs/filter/criteria-based-filter-form.tsx
```diff
@@ -0,0 +1,38 @@
...
+    [ENERGY_SOURCE]: yup.string().nullable(),
+    ...filterPropertiesYupSchema,
+});
+
+export const criteriaBasedFilterEmptyFormData = getCriteriaBasedFormData(null, {
+    [ENERGY_SOURCE]: null,
+    [FILTER_PROPERTIES]: [],
+});
+
+function CriteriaBasedFilterForm() {
```
klesaulnier, line 26:
```suggestion
export const CriteriaBasedFilterForm: FunctionComponent = () => {
```

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
```diff
@@ -0,0 +1,133 @@
...
+    // default values are set via reset when we fetch data
+    const formMethods = useForm({
+        resolver: yupResolver(formSchema),
+    });
+
+    const { reset, setValue } = formMethods;
+
+    // Fetch the filter data from back-end if necessary and fill the form with it
+    useEffect(() => {
+        if (id && open) {
```
klesaulnier, line 54: are there any case when open or id are falsy ?

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
```diff
@@ -0,0 +1,133 @@
...
+                },
+                null,
+                intl,
+                onClose
+            );
+        },
+        [id, intl, onClose, snackError]
+    );
+
+    const checkName = (isValid, newName) => {
```
klesaulnier:
```suggestion
    const handleNameChange = (isValid, newName) => {
```

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
```diff
@@ -0,0 +1,133 @@
...
+        setValue(NAME, newName);
+    };
+
+    return (
+        <CustomMuiDialog
+            open={open}
+            onClose={onClose}
+            onSave={onSubmit}
+            schema={formSchema}
+            methods={formMethods}
```
klesaulnier:
```suggestion
            formSchema={formSchema}
            formMethods={formMethods}
```

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
```diff
@@ -0,0 +1,133 @@
...
+            titleId={titleId}
+            removeOptional={true}
+            disabledSave={!isNameValid}
+        >
+            <NameWrapper
+                titleMessage="nameProperty"
+                initialValue={name}
+                contentType={ElementType.FILTER}
+                handleNameValidation={checkName}
+            >
```
klesaulnier: Maybe to do later, but this component is currently not handled by RHF, any reason why ?

src/components/dialogs/filter/explicit-naming-filter-form.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    ),
+        }),
+};
+
+export const explicitNamingFilterEmptyFormData = {
+    [FILTER_EQUIPMENTS_ATTRIBUTES]: [],
+};
+
+const defaultRowData = {
+    [EQUIPMENT_ID]: null,
```
klesaulnier:
```suggestion
    [EQUIPMENT_ID]: "",
```
empty value for string inputs should be empty string

src/components/dialogs/filter/explicit-naming-filter-form.tsx
```diff
@@ -0,0 +1,162 @@
...
+            </Grid>
+            <Grid item xs={12}>
+                <CustomAgGridTable
+                    name={FILTER_EQUIPMENTS_ATTRIBUTES}
+                    columnDefs={columnDefs}
+                    defaultRowData={defaultRowData}
+                    defaultEmptyRowsNumber={3}
+                    pagination={true}
+                    paginationPageSize={100}
+                    minNumberOfRows={3}
```
klesaulnier: I'm not sure it's used anywhere

src/components/dialogs/filter/filter-properties.tsx
```diff
@@ -0,0 +1,213 @@
...
+            return Promise.reject(
+                'Study entry could not be found in metadatas'
+            );
+        }
+
+        return studyMetadata?.predefinedEquipmentProperties?.substation;
+    });
+}
+
+function FilterProperties() {
```
klesaulnier, line 121:
```suggestion
export const FilterProperties: FunctionComponent = (
```

src/components/dialogs/filter/filter-properties.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+export const FILTER_PROPERTIES = 'freeProperties';
+
+function propertyValuesTest(
+    values: (string | undefined)[] | undefined,
+    context: yup.TestContext<yup.AnyObject>,
+    doublePropertyValues: boolean
+) {
+    // with context.from[length - 1], we can access to the root fields of the form
+    const rootLevelForm = context!.from![context!.from!.length - 1];
```
klesaulnier: since context seems to be mandatory, we don't need "!"

src/components/dialogs/filter/filter-properties.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+    const isForLineOrHvdcLine =
+        watchEquipmentType === Line.type || watchEquipmentType === Hvdc.type;
+
+    const isForSubstation = watchEquipmentType === Substation.type;
+
+    function addNewProp() {
+        if (isForLineOrHvdcLine) {
+            append({
+                [PROPERTY_NAME]: '',
```
klesaulnier:
```suggestion
                [PROPERTY_NAME]: null,
```
=> it's a select input, default empty value should be null to prevent "delete button" from appearing

src/components/dialogs/filter/filter-properties.tsx
```diff
@@ -0,0 +1,213 @@
...
+                [PROPERTY_VALUES_1]: [],
+                [PROPERTY_VALUES_2]: [],
+            });
+        } else {
+            append({ [PROPERTY_NAME]: '', [PROPERTY_VALUES]: [] });
```
klesaulnier:
```suggestion
            append({ [PROPERTY_NAME]: null, [PROPERTY_VALUES]: [] });
```

src/components/dialogs/filter/filter-properties.tsx
```diff
@@ -0,0 +1,213 @@
...
+    const valuesFields = isForLineOrHvdcLine
+        ? [
+              { name: PROPERTY_VALUES_1, label: 'PropertyValues1' },
+              { name: PROPERTY_VALUES_2, label: 'PropertyValues2' },
+          ]
+        : [{ name: PROPERTY_VALUES, label: 'PropertyValues' }];
+
+    useEffect(() => {
+        fetchPredefinedProperties()
+            .then((p) => setFieldProps(p))
```
klesaulnier, line 165: it could be nice to type "p" and "fieldProps", it would prevent "any" from being used in "filter-property.tsx"

src/components/dialogs/filter/filter-properties.tsx
```diff
@@ -0,0 +1,213 @@
...
+                            predefined={fieldProps}
+                            handleDelete={remove}
+                        />
+                    </ListItem>
+                ))}
+                <Grid item>
+                    <Button
+                        fullWidth
+                        startIcon={<AddIcon />}
+                        onClick={() => addNewProp()}
```
klesaulnier:
```suggestion
                        onClick={addNewProp}
```

src/components/dialogs/filter/filter-property.tsx
```diff
@@ -0,0 +1,103 @@
...
+export const PROPERTY_VALUES_2 = 'prop_values2';
+
+interface FilterPropertyProps {
+    index: number;
+    valuesFields: Array<{ name: string; label: string }>;
+    handleDelete: Function;
+    predefined: any;
+}
+
+function FilterProperty(props: FilterPropertyProps) {
```
klesaulnier, line 23:
```suggestion
export const FilterProperty: FunctionComponent<FilterPropertyProps> = (
```

src/components/dialogs/filter/filter-property.tsx
```diff
@@ -0,0 +1,103 @@
...
+export const PROPERTY_NAME = 'name_property';
+export const PROPERTY_VALUES = 'prop_values';
+export const PROPERTY_VALUES_1 = 'prop_values1';
+export const PROPERTY_VALUES_2 = 'prop_values2';
+
+interface FilterPropertyProps {
+    index: number;
+    valuesFields: Array<{ name: string; label: string }>;
+    handleDelete: Function;
+    predefined: any;
```
klesaulnier, line 20: can we type this ?

src/components/dialogs/filter/filter-property.tsx
```diff
@@ -0,0 +1,103 @@
...
+function FilterProperty(props: FilterPropertyProps) {
+    const { setValue } = useFormContext();
+
+    const watchName = useWatch({
+        name: `${CRITERIA_BASED}.${FILTER_PROPERTIES}[${props.index}].${PROPERTY_NAME}`,
+    });
+
+    const predefinedNames = useMemo(() => {
+        return Object.keys(props.predefined ?? []).sort();
+    }, [props.predefined]);
```
klesaulnier, line 32: destructure props at the beginning of the component instead of using props.* everywhere

src/components/dialogs/filter/filter-property.tsx
```diff
@@ -0,0 +1,103 @@
...
+                                    label={option}
+                                    {...getTagsProps({ index })}
+                                />
+                            ))
+                        }
+                        // we have to fill these fields even if we don't use it because we use strict TS
+                        previousValue={undefined}
+                        formProps={undefined}
+                        allowNewValue={undefined}
+                        onChangeCallback={undefined}
```
klesaulnier, line 90: I think it's cleaner to add default values to AutocompleteInput properties

previousValue = undefined,
allowNewValue = undefined,
(...)

src/components/dialogs/filter/filter-property.tsx
```diff
@@ -0,0 +1,103 @@
...
+            {props.valuesFields.map((valuesField) => (
+                <Grid item xs key={valuesField.name}>
+                    <AutocompleteInput
+                        name={`${CRITERIA_BASED}.${FILTER_PROPERTIES}[${props.index}].${valuesField.name}`}
+                        label={valuesField.label}
+                        options={predefinedValues}
+                        freeSolo
+                        forcePopupIcon
+                        multiple
+                        renderTags={(val: any, getTagsProps: any) =>
```
klesaulnier:
```suggestion
                        renderTags={(val: string[], getTagsProps: any) =>
```

src/components/dialogs/filter/filters-save.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const saveExplicitNamingFilter = (
+    tableValues,
+    isFilterCreation,
+    equipmentType,
+    name,
+    id,
+    setCreateFilterErr,
+    activeDirectory,
+    intl,
```
klesaulnier: unused param

src/components/utils/rhf-inputs/ag-grid-table-rhf/custom-ag-grid-table.js
```diff
@@ -69,6 +69,7 @@ export const CustomAgGridTable = ({
     name,
     columnDefs,
     defaultRowData,
+    defaultEmptyRowsNumber,
```
klesaulnier: I'm not sure if it's a behaviour we should keep

klesaulnier (changes requested):
Some bugs : 
- 2 columns can be moved and hidden in AGGrid table
- typing text other than number in "distributionKey" causes weird behavior. You can type anything but it will be removed on lost focus
- AGGrid table is displayed even if no type is selected

YenguiSeddik (commented):
Some small remarques after testing
- When editing criteria based filter (with filled fields) and we change equipments type, there is no confirmation popup and fields are not initialized. is this the wanted behavior ?
- When selecting "Explicit naming", the table is displayed even there is no equipment type
- When adding rows in explicit naming filter for generators ot load, and forget to add a key distribution for row while other have one, you get an error (which expected), but if you change equipment type for somthing that does not require a distribution key, you still get the error. Maybe initilize the table (or at least distribution key) when changing equipment type

src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js
flomillot: What do you mean ? As it was done before, in the defaultValues ?

src/components/dialogs/filter/criteria-based-filter-form.tsx
flomillot: Why ?

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
flomillot: I don't think so, I just kept the previous code, and better to have too much safety checks than not enough

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
flomillot: I didn't know which one to keep, this method had several different names, so I put handleNameChange everywhere right ?

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
flomillot: CustomMuiDialog is not my component, but I agree it's clearer, so I will change it

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
flomillot: The US was big enough so we limited the changes to what we need, but indeed we should

src/components/dialogs/filter/explicit-naming-filter-form.tsx
flomillot: Ok, any reason ?

src/components/dialogs/filter/explicit-naming-filter-form.tsx
flomillot: It was added in the contingency list so I added it too : https://github.com/gridsuite/gridexplore-app/blob/3f1201e0a56a5997990bc3ec24556cb30bfe9f4d/src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js#L117
I guess it's used in AgGridReact no ?

src/components/dialogs/filter/filter-properties.tsx
flomillot: Why ?

src/components/dialogs/filter/filter-properties.tsx
flomillot: I think I just kept the previous default value, you're right I will change it

src/components/dialogs/filter/filter-properties.tsx
flomillot: Tbh I don't know what's inside and only reused the previous code with the `predefined` field.
To remove `any` from this file, the job actually need to be done here : https://github.com/gridsuite/gridexplore-app/blob/3f1201e0a56a5997990bc3ec24556cb30bfe9f4d/src/components/dialogs/filter/filter-properties.tsx#L109
Should I go in the back end to check what it contains to type it correctly ?

src/components/dialogs/filter/filter-property.tsx
flomillot: Why²

src/components/dialogs/filter/filter-property.tsx
flomillot: Same as above, it's black box for me, but I can go in the back end check what it contains if you want to

src/components/dialogs/filter/filter-property.tsx
flomillot: Is it mandatory ? I prefer much more when it's not desctructured then I easily know it's a prop.
Maybe I can ask on slack ?

src/components/dialogs/filter/filter-property.tsx
flomillot: But we said we don't change inputs until it's migrated in commons ui

flomillot: I note it in the associated TS maybe ?

src/components/dialogs/filter/filter-property.tsx
flomillot: You're right i forgot to type that

src/components/utils/rhf-inputs/ag-grid-table-rhf/custom-ag-grid-table.js
flomillot: What do you suggests instead ?

flomillot:
> Some small remarques after testing
> 
> * When editing criteria based filter (with filled fields) and we change equipments type, there is no confirmation popup and fields are not initialized. is this the wanted behavior ?
> * When selecting "Explicit naming", the table is displayed even there is no equipment type
> * When adding rows in explicit naming filter for generators ot load, and forget to add a key distribution for row while other have one, you get an error (which expected), but if you change equipment type for somthing that does not require a distribution key, you still get the error. Maybe initilize the table (or at least distribution key) when changing equipment type

- Actually it was the bug you fixed in this PR https://github.com/gridsuite/gridexplore-app/pull/251#pullrequestreview-1545224034, I updated the branch so now it works
- Indeed I forgot the change that, I will
- Oh yeah I see, I will fix it, thanks

src/components/dialogs/filter/filters-save.js
```diff
@@ -0,0 +1,71 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+import { createFilter, saveFilter } from '../../../utils/rest-api';
+import { FilterType } from '../../../utils/elementType';
+import { frontToBackTweak } from './criteria-based-filter-dialog-utils';
+import { NAME } from '../../utils/field-constants';
+
+export const saveExplicitNamingFilter = (
```
YenguiSeddik, line 14: I don't think we need this method, we can use createFilter  in create-filter-dialog.js and  saveFilter in explicit-naming-filter-edition-dialog.js directly

flomillot: I kept the previous code the more I can (it's what tech lead usually told me to do)

flomillot: I kept the previous code the more I can (it's what tech lead usually told me to do)

flomillot: I kept the previous code the more I can (it's what tech lead usually told me to do)

flomillot:
> Some bugs :
> 
> * 2 columns can be moved and hidden in AGGrid table
> * typing text other than number in "distributionKey" causes weird behavior. You can type anything but it will be removed on lost focus
> * AGGrid table is displayed even if no type is selected

Fixed, except for the letters behaviour, I'm waiting for PO answer

src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js
klesaulnier: if we need 3 empty rows when opening the dialog, it should be dealt with "emptyFormData" which should start with a table having 3 empty rows

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
klesaulnier: I think it's nice to change names if they make the code clearer, but it can be done later

src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.js
flomillot: Indeed if we delete all the rows, with my code they come back...
OK I revert this part.

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
flomillot: I mean several names was used for the same method, I took the one the most used. I can change it's fast, so I put handleNameChange everywhere ?

src/components/dialogs/filter/explicit-naming-filter-form.tsx
klesaulnier: I don't think it's used anywhere, plus, you can actually remove rows and validate while having less than 3 rows

src/components/dialogs/filter/filter-properties.tsx
klesaulnier: Yes I think it could be nice to do so

src/components/dialogs/filter/filter-property.tsx
klesaulnier: I think it could be nice to do so, we will need to do it one day anyway

src/components/dialogs/filter/explicit-naming-filter-form.tsx
flomillot: Ok done

src/components/dialogs/filter/filter-property.tsx
klesaulnier: After some researches, it's up to personal preferences so we can leave it like that, except if we want everyone to code the same way

src/components/dialogs/filter/explicit-naming-filter-edition-dialog.js
klesaulnier: the code structure around name change handling is a bit weird, it's called "handleNameValidation" in NameWrapper, whereas it's not really supposed to validate anything but communicate its value to its parent

so I would leave it like that until we fix this component (NameWrapper)

flomillot: Yeah indeed, I won't change NameWrapper, but just change to handleNameChange because it does not check anything and it's clearer like that.

src/components/dialogs/filter/explicit-naming-filter-form.tsx
flomillot: Removed

src/components/dialogs/filter/filter-property.tsx
flomillot: I did it

klesaulnier (approved):
test OK
code OK

YenguiSeddik (approved):
code: ok
test: ok

PR #268, opened 2023-08-09
Migrate the RHF inputs from gridstudy-app to commons-ui

flomillot:


decs.d.ts
```diff
@@ -15,4 +20,104 @@ declare module '@gridsuite/commons-ui' {
...
+
+    interface AutocompleteInputProps
+        extends Omit<
+            AutocompleteProps<
+                string,
+                boolean | undefined,
+                boolean | undefined,
+                boolean | undefined
+            >,
+            'renderInput' // we already defined it in our custom Autocomplete
```
klesaulnier: omit "onChange" and "value" as well

decs.d.ts
```diff
@@ -15,4 +20,104 @@ declare module '@gridsuite/commons-ui' {
...
+        > {
+        name: string;
+        label?: string;
+        options: Array<any>; // https://mui.com/material-ui/react-autocomplete/#options-structure
```
klesaulnier: I think we don't need it here since it's already defined in AutocompleteProps

decs.d.ts
```diff
@@ -15,4 +20,104 @@ declare module '@gridsuite/commons-ui' {
...
+        outputTransform?: (value: string) => string;
+        inputTransform?: (value: string) => string;
+        readOnly?: boolean;
+        previousValue?: string;
+        allowNewValue?: boolean;
+        onChangeCallback?: () => void;
+        formProps?: TextFieldProps;
+    }
+
+    export function AutocompleteInput(props: AutocompleteInputProps);
```
klesaulnier:
```suggestion
    export function AutocompleteInput(props: AutocompleteInputProps): ReactNode;
```

same for all exports, otherwise it returns any

decs.d.ts
```diff
@@ -15,4 +20,104 @@ declare module '@gridsuite/commons-ui' {
...
+
+    interface ErrorInputProps {
+        name: string;
+        InputField?: ReactElement;
+    }
```
klesaulnier:
```suggestion
    interface ErrorInputProps {
        name: string;
        InputField?: ReactNode;
    }
```

now FC returns a ReactNode instead of ReactElement, I think we should use ReactNode as well

klesaulnier (approved):

decs.d.ts
flomillot: Actually after few research I think we should not, in our case it can only be a ReactElement

decs.d.ts
flomillot: Ok I will type the return value.
But they are all ReactElement, why using a wider type ?

decs.d.ts
flomillot: [ReactNode is wider, it can be text, number, boolean, null, undefined, a portal, a ReactElement, or an array of ReactNodes. It represents anything that React can render.](https://stackoverflow.com/a/72353143)

klesaulnier (approved):

thangqp (approved):
OK

PR #272, opened 2023-08-25
Various fixes on explicit naming filters and contengency lists

flomillot:
In AG Grid tables for explicit naming filters and contengency lists :
- for checkboxes, use same blue color than in directory view and buttons
- dont use a bg color for selected rows
- remove the extra column on the left of the checkboxes
- add a loader, for huge tables
- when we add a new row, scroll to last page/row
- supress the square on selected cells
- fix ChipsArrayEditor width in the table
- improve performances when we delete all the rows
- improve performances when we add big CSV


src/setupTests.js
```diff
@@ -6,3 +6,12 @@
...
 global.IS_REACT_ACT_ENVIRONMENT = true;
+
+// make crypto.randomUUID() available in tests
+import crypto from 'crypto';
+
+window.crypto = {
+    randomUUID: function () {
+        return crypto.randomUUID();
+    },
+};
```
Walid-Sahnoun: add empty line

src/setupTests.js
```diff
@@ -6,3 +6,12 @@
  */
 
 global.IS_REACT_ACT_ENVIRONMENT = true;
+
```
Walid-Sahnoun: replace global.IS_REACT_ACT_ENVIRONMENT = true; under import

YenguiSeddik (approved):

Walid-Sahnoun (approved):
code and test ok

PR #277, opened 2023-09-04
Auto-Increment and Error Messaging for Duplicate Contingency List Names in Script

achour94:
Implement Automatic Incrementation and Error Message for existing filters contingency list Name when Copied to Script


src/components/dialogs/name-wrapper.js
```diff
@@ -104,6 +105,14 @@ const NameWrapper = ({
         );
     };
 
+    //check if name already exist if initialValue is not empty
+    useEffect(() => {
+        if (initialValue !== '') {
+            updateFormState(initialValue);
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
jonenst: forbidden in our code base :)

souissimai (approved):
Code Ok
Test Ok

src/components/dialogs/name-wrapper.js
```diff
@@ -104,6 +98,14 @@ const NameWrapper = ({
         );
     };
 
+    //check if name already exist if initialValue is not empty
+    useEffect(() => {
+        if (initialValue !== '') {
+            updateFormState(initialValue);
+            setValue(initialValue);
```
jonenst, line 105: Does this effect run multiple times ? If not, there's not need for setValue as initialValue is passed to the useState() ?

jonenst: It's a bad name "initialValue" if it triggers additional behaviors when it changes. In react the convention is to only use the value for the first render

achour94: yes, it willl trigger the useeffect twice in the case we will get the name from the backend

src/components/dialogs/copy-to-script-dialog.js
```diff
@@ -57,23 +76,29 @@ const CopyToScriptDialog = ({
...
-                    autoFocus
-                    value={newNameValue}
-                    required={true}
-                    onChange={updateNameValue}
-                />
+                {loading ? (
+                    <CircularProgress />
+                ) : (
+                    <NameWrapper
+                        initialValue={generatedName}
```
jonenst, line 83: I would fix the initialValue behavior and restore the value={...} and onChange{...} prop

src/components/dialogs/copy-to-script-dialog.js
```diff
@@ -26,17 +27,25 @@ import PropTypes from 'prop-types';
...
+    onGenerateNameError,
     open,
     onClose,
     onClick,
     currentName,
     title,
 }) => {
-    const [newNameValue, setNewNameValue] = React.useState(currentName);
+    const [generatedName, setGeneratedName] = useState('');
+    const [newNameValue, setNewNameValue] = useState(currentName);
```
SlimaneAmar, line 40: Why we do use generatedName and newNameValue
Maybe one is enough

achour94: because in that case (initialValue={newNameValue}) the useEffect with initialValue as a dependency will be triggered each time newNameValue is changed

achour94:
replaced with https://github.com/gridsuite/gridexplore-app/pull/285

PR #279, opened 2023-09-08
Replace useNameCheck and NameWrapper by a RHF component

flomillot:
Create a new react-hook-form input component to replace both useNameCheck and NameWrapper.
useNameCheck is still used by some old components btw.
Some bugs fixes, cleaning and refactoring.

PR #281, opened 2023-09-11
filters forms: add a space between filter name and filter type

dbraquart:
no need this PR anymore (fixed it in https://github.com/gridsuite/gridexplore-app/pull/279)


flomillot (changes requested):
I've already fixed it in https://github.com/gridsuite/gridexplore-app/pull/279
List of fixed bugs are in TS 2368

PR #282, opened 2023-09-14
Encode element name before api call.

AAJELLAL:


souissimai (approved):
code and test ok

src/utils/rest-api.js
```diff
@@ -456,10 +456,13 @@ export function duplicateCase(
 }
 
 export function elementExists(directoryUuid, elementName, type) {
+    const elementNameEncoded = new URLSearchParams({
+        elementName: elementName,
```
anistouri, line 460: you can simply use this shortcut using destructuring
`new URLSearchParams({ elementName }).toString()`

anistouri (approved):

src/utils/rest-api.js
AAJELLAL: Yes indeed, but i prefer keep it this way if we ever need to add anything else.

PR #287, opened 2023-09-26
Use params default values when creating study

YenguiSeddik:


src/components/dialogs/create-study-dialog/create-study-dialog.js
```diff
@@ -126,6 +127,21 @@ const CreateStudyDialog = ({ open, onClose, providedExistingCase }) => {
...
+                                : param.defaultValue,
+                        })
+                    );
+
+                    const currentParametersObj = Object.assign(
+                        {},
+                        ...currentParameters
+                    );
+
+                    setValue(CURRENT_PARAMETERS, currentParametersObj);
```
ghelloumi: you can use reduce if you want

const currentParametersObj = result.parameters.reduce((obj, param) => {
  obj[param.name] = Array.isArray(param.defaultValue) ? param.defaultValue.toString() : param.defaultValue;
  return obj;
}, {});

setValue(CURRENT_PARAMETERS, currentParametersObj);

ghelloumi (approved):

src/components/dialogs/create-study-dialog/create-study-dialog.js
```diff
@@ -49,6 +50,30 @@ import PrefilledNameInput from '../commons/prefilled-name-input';
 
 const STRING_LIST = 'STRING_LIST';
 
+function customizeCurrentParameters(params) {
+    return params.reduce((obj, parameter) => {
+        // we check if the parameter is for extension, if it is, we select all possible values by default.
```
etiennehomer: we check if the parameter is for extensions. If so, we select all possible values by default.

etiennehomer, line 55: we check if the parameter is for extensions. If so, we select all possible values by default.

etiennehomer: we check if the parameter is for extensions. If so, we select all possible values by default.

src/components/dialogs/create-study-dialog/create-study-dialog.js
```diff
@@ -49,6 +50,30 @@ import PrefilledNameInput from '../commons/prefilled-name-input';
...
+        // the only way for the moment to check if the parameter is for extension, is by checking his name.by def
```
etiennehomer: Remove "by def" nope ?

etiennehomer: And you can keep a TODO
TODO: implement a cleaner way to determine the extensions field

etiennehomer (approved):

sBouzols (approved):

PR #289, opened 2023-10-05
add a itemMenu to download selected cases from gridexplore

jamal-khey:


AbdelHedhili (changes requested):
the menu should close once you click once the download starts

src/components/menus/content-contextual-menu.js
```diff
@@ -380,6 +383,43 @@ const ContentContextualMenu = (props) => {
...
+            const name = await getOriginalName(element.elementUuid);
+            const blob = await result.blob();
+            const href = window.URL.createObjectURL(blob);
+            const a = linkRef.current;
+            a.download = name;
+            a.href = href;
+            a.click();
+            a.href = '';
+        });
+    }, [selectedElements]);
```
AbdelHedhili: It would be nice to use the same way as we already do in "network-modification-tree-pane.js" in gridstudy-app to download the case

AbdelHedhili (changes requested):

src/components/menus/content-contextual-menu.js
jamal-khey: Hello ,
in my case , i don't have a new pop-up, so i don't have anything to close. 
when the user click on download , it automatically download all selected cases

AbdelHedhili: Not talking about pop-up i'm talking about the menu, when you click on download the menu shoudn't stay open

jamal-khey: sorry for the delay.
yes you are right , the menue does not close when we click on download.  i fixed it.

AbdelHedhili (changes requested):
Change to code to use Iframes the same way we do it in gridstudy-app to download a case

src/components/menus/content-contextual-menu.js
jamal-khey: @AbdelHedhili for some reason in my case it doesn't work in my case because i recieve a blob from the API. i managed to download it but i didn't find a way to give it a name.
in study server the API return a url. and not a blob 

from what i understod from the documentation, iframe are used to ["to display a web page within a web page"](https://www.w3schools.com/html/html_iframe.asp) .

also most [examples](https://www.w3docs.com/learn-javascript/javascript-blob.html) i found on internet that save a blob are using <a> or  tag 

if this remark is not bloking for you i suggest to keep <a> tag

src/utils/rest-api.js
```diff
@@ -483,6 +483,24 @@ export function getNameCandidate(directoryUuid, elementName, type) {
     });
 }
 
+/**
```
dbraquart, line 486: you should specify everywhere below we are talking about a case, because this function is case-related.
ex: getCaseOriginalName(caseUuuid) ... etc

src/components/menus/content-contextual-menu.js
AbdelHedhili: ok let's not overcomplicate things since it works that way let's keep it.

src/components/menus/content-contextual-menu.js
```diff
@@ -380,6 +383,43 @@ const ContentContextualMenu = (props) => {
...
 
+    const allowsDownloadCase = useCallback(() => {
+        //if selectedElements contains atleast one case
+        return selectedElements.some(
+            (element) => element.type === ElementType.CASE
+        );
+    }, [selectedElements]);
+
+    /**
+     * Downloads the selected cases as a file.
```
dbraquart, line 411:
```suggestion
     * Downloads each selected case as a file.
```

src/components/menus/content-contextual-menu.js
```diff
@@ -380,6 +383,43 @@ const ContentContextualMenu = (props) => {
...
+     * @name handleDownloadCase
+     * @returns {void}
+     */
+    const handleDownloadCase = useCallback(() => {
+        //for each selectedElements , filter cases and return their uuid and subtype
+        const elementsToDownload = selectedElements
+            .filter((element) => element.type === ElementType.CASE)
+            .map((element) => ({
+                elementUuid: element.elementUuid,
+                subtype: element.subtype,
```
dbraquart: it looks like subtype is never used

src/components/menus/content-contextual-menu.js
```diff
@@ -380,6 +383,43 @@ const ContentContextualMenu = (props) => {
         );
     }, [isUserAllowed, selectedElements]);
 
+    const allowsDownloadCase = useCallback(() => {
+        //if selectedElements contains atleast one case
```
dbraquart:
```suggestion
        //if selectedElements contains at least one case
```

dbraquart (changes requested):
code: a few remarks
tests: ok

src/utils/rest-api.js
jamal-khey: done

src/components/menus/content-contextual-menu.js
jamal-khey: done

src/components/menus/content-contextual-menu.js
jamal-khey: done

src/components/menus/content-contextual-menu.js
jamal-khey: done

AbdelHedhili (approved):

src/utils/rest-api.js
```diff
@@ -483,6 +483,24 @@ export function getNameCandidate(directoryUuid, elementName, type) {
...
+ * Retrieves the original name of a case using its UUID.
+ * @param {string} caseUuid - The UUID of the element.
+ * @returns {Promise<string|boolean>} - A promise that resolves to the original name of the case if found, or false if not found.
+ */
+export function getCaseOriginalName(elementUuid) {
```
dbraquart: sorry but JsDoc above is not good : caseUuid vs elementUuid

dbraquart (changes requested):

src/utils/rest-api.js
jamal-khey: sorry , i corrected it

dbraquart (approved):

PR #293, opened 2023-10-18
change appearance validate and cancel button

anistouri:


YenguiSeddik (approved):

src/components/utils/popup-confirmation-dialog.js
```diff
@@ -26,9 +27,9 @@ const PopupConfirmationDialog = ({
                 </DialogContentText>
             </DialogContent>
             <DialogActions>
-                <Button onClick={() => setOpenConfirmationPopup(false)}>
+                <CancelButton onClick={() => setOpenConfirmationPopup(false)}>
                     <FormattedMessage id="cancel" />
```
flomillot: There is already `<FormattedMessage id="cancel" /> `in `CancelButton` no ?

src/components/dialogs/delete-dialog.js
```diff
@@ -31,6 +31,9 @@ const styles = {
     tooltip: {
         maxWidth: '1000px',
     },
+    deleteButtonTextColor: (theme) => ({
+        color: theme.palette.text.primary,
```
flomillot: Don't use `sx` and use `color` instead

src/components/dialogs/replace-with-script-dialog.js
```diff
@@ -48,10 +49,12 @@ const ReplaceWithScriptDialog = ({ id, open, onClose, onClick, title }) => {
...
-                <Button onClick={handleClose} variant="outlined">
-                    <FormattedMessage id="cancel" />
-                </Button>
-                <Button onClick={handleClick}>
+                <CancelButton
+                    onClick={handleClose}
+                    variant="outlined"
+                    withCustomColor={true}
+                />
+                <Button onClick={handleClick} color={'customButton'}>
```
flomillot, line 57: Validate in grey too ?

src/components/dialogs/delete-dialog.js
```diff
@@ -210,14 +210,17 @@ const DeleteDialog = ({
...
+                <Button
+                    onClick={handleClick}
+                    disabled={loadingState}
+                    color={'customButton'}
                 >
-                    <FormattedMessage id="cancel" />
-                </Button>
-                <Button onClick={handleClick} disabled={loadingState}>
                     {(loadingState && <CircularProgress size={24} />) || (
                         <FormattedMessage id="delete" />
```
flomillot, line 225: In the US they don't specify to change the delete button. If so, we should do it everywhere no ?

src/components/dialogs/replace-with-script-dialog.js
anistouri: When the cancel button is customized, the secondary button should be white on the dark theme and black on the light theme

src/components/dialogs/delete-dialog.js
anistouri: After discussion with the PO, we are implementing it here and this component is generic and used wherever there is a deletion

flomillot (approved):

PR #295, opened 2023-11-02
Expert filter improvements

antoinebhs:
* Isolate country value editor and energy source value editor in separated components. Energy source value editor is generalized in a "translated" value component that can be reused if other selectable inputs need to be translated.
* Don't show an error when adding a new rule. Only show an error when the number of rules is the same.


src/components/utils/rqb-inputs/translated-value-editor.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+    return (
+        <MaterialValueEditor
+            {...props}
+            values={props.values?.map((v) => {
+                return {
+                    name: v.name,
+                    label: intl.formatMessage({ id: v.label }),
+                };
+            })}
```
thangqp: useMemo

src/components/utils/dialog-utils.js
```diff
@@ -34,3 +38,20 @@ export const toFloatOrNullValue = (value) => {
...
+                    getComputedLanguage(languageLocal).substr(0, 2))
+            );
+        } catch (error) {
+            // fallback to English if no localized list is found
+            return require('localized-countries')(
+                require('localized-countries/data/en')
+            );
+        }
+    }, [languageLocal]);
+};
```
thangqp: In my opinion, can try to use the hook in gridstudy-app : src/components/utils/localized-countries-hook.js
We can duplicate this hook in gridexplore-app. Eventually, we have already a POC to merge between two apps in the next year

src/components/utils/rhf-inputs/select-inputs/countries-input.js
```diff
@@ -1,25 +1,10 @@
...
-                    getComputedLanguage(languageLocal).substr(0, 2))
-            );
-        } catch (error) {
-            // fallback to english if no localised list found
-            return require('localized-countries')(
-                require('localized-countries/data/en')
-            );
-        }
-    }, [languageLocal]);
+    const countriesListCB = useCountriesListCB();
```
thangqp: In case of using src/components/utils/localized-countries-hook.js by duplication, do not need refactor this component. We can do it later..

src/components/utils/rqb-inputs/custom-react-query-builder.tsx
```diff
@@ -41,14 +42,15 @@ const CustomReactQueryBuilder = (props: CustomReactQueryBuilderProps) => {
 
     const handleQueryChange = useCallback(
         (newQuery: RuleGroupTypeAny) => {
-            const hasChanged =
-                formatQuery(
-                    getValues(EXPERT_FILTER_QUERY),
-                    'json_without_ids'
-                ) !== formatQuery(newQuery, 'json_without_ids');
+            const oldQuery = getValues(EXPERT_FILTER_QUERY);
```
thangqp: props.name ???

thangqp (approved):

src/components/utils/rhf-inputs/select-inputs/countries-input.js
antoinebhs: I don't think duplicating the code is good in the current context as we might merge the gridstudy and gridexplore soon and duplicating this code might lead to divergence. We'll update this when the merge is done.

src/components/utils/dialog-utils.js
antoinebhs: I don't think duplicating the code is good in the current context as we might merge the gridstudy and gridexplore soon and duplicating this code might lead to divergence. We'll update this when the merge is done.

thangqp: OK

src/components/utils/rhf-inputs/select-inputs/countries-input.js
thangqp: OK

PR #296, opened 2023-11-03
Remove the need for a secure context for the app

antoinebhs:
Issue: we need a secure context to run the gridexplore-app because it's required by crypto.randomUUID() (see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID). This leads to some extra steps to run a k8s cluster locally and forces us to use a secure context for the app while it does not seem necessary given that we use crypto.randomUUID() to generate row ids in gridexplore.

We have some options to solve this issue:
- generate a random UUID with methods that do not need a secure context: 
    - use [uuid](https://www.npmjs.com/package/uuid) package already used in *commons-ui* (https://github.com/gridsuite/commons-ui/blob/b82955b4bb1901655176175f3bc2b508e0c7c154/src/components/ReportViewer/log-report.js#L14)
    - use Math.random() to avoid adding a package, this is not the recommended approach from what I've read online because it does not provide good uniqueness guarantees
- update the local k8s deployment instructions to somehow run the app in a secure context (self-signed certificate? proxy?)


Tristan-WorkGH (approved):

sBouzols (approved):
Code Review OK

PR #299, opened 2023-11-08
add outlined on SubmitButton

anistouri:


src/components/app-wrapper.js
```diff
@@ -128,11 +124,27 @@ const darkTheme = createTheme({
     },
 });
 
+const customDarkTheme = createTheme(darkTheme, {
```
sBouzols: I would prefer this solution here : https://stackoverflow.com/a/69450044/5359613
to avoid theme customisation in darkTheme fro some, then in customDarkTheme for others
```
let defaultTheme = createTheme();
const lightTheme = createTheme(defaultTheme, {...});
const darkTheme = createTheme(defaultTheme, {...});
```

decs.d.ts
```diff
@@ -137,17 +142,13 @@ declare module '@gridsuite/commons-ui' {
...
+    export const SubmitButton: FunctionComponent<ButtonProps>;
 
-    export const CancelButton: FunctionComponent<{
-        onClick: (() => void) | ((event: React.MouseEvent) => void);
-        disabled?: boolean;
-        variant?: string;
-        withCustomColor?: boolean;
-    }>;
+    interface CancelButtonProps extends ButtonProps {
+        color?: string;
```
sBouzols, line 148: ButtonProps has already a color prop no ?

src/components/app-wrapper.js
```diff
@@ -47,9 +49,6 @@ import { PARAM_THEME } from '../utils/config-params';
 const lightTheme = createTheme({
     palette: {
         mode: 'light',
-        customButton: {
-            main: '#000000DE',
-        },
```
sBouzols, line 52: hum I don't get it, those custom colors were already there ?

sBouzols (changes requested):

src/components/app-wrapper.js
anistouri: Yes, that was the first solution I had implemented

decs.d.ts
anistouri: The prop color of ButtonProps accepts only these values: 'inherit' | 'primary' | 'secondary' | 'success' | 'error' | 'info' | 'warning'

src/components/app-wrapper.js
anistouri: I relied on the MUI documentation in the 'Using Theme Options to Define Other Options' section'
https://mui.com/material-ui/customization/theming/#api

sBouzols: You apply theme composition here but we just want a composition between the default theme and our custom theme
* before : defaultTheme + lightTheme|darkTheme
* with your change : defaultTheme + lightTheme|darkTheme + customDarkTheme|customLightTheme

this is not necessary I think,  lightTheme|darkTheme are already our custom themes

decs.d.ts
sBouzols: https://mui.com/material-ui/api/button/#Button-prop-color
I see that it accepts string as well

src/components/app-wrapper.js
sBouzols: ok

src/components/app-wrapper.js
anistouri: You cannot pass defaultTheme as a parameter to createTheme in this way, it will not work because createTheme does not expect to receive an existing theme as the first argument, but rather a configuration object or nothing at all. Therefore, one can use the defaultTheme constant directly without passing it as a parameter. This works, but you need to set the mode so that the text color adapts to the theme; otherwise, it will always be black (the default color). For example:
` const defaultTheme = createTheme({ palette: { mode: 'dark', }, });`

anistouri: Done

decs.d.ts
anistouri: Like we discussed earlier, I kept that

sBouzols (approved):
Code Review OK

PR #310, opened 2023-12-08
feat(authentication): handle id_token renewal

TheMaskedTurtle:
The aim is to handle id_token in our client code as it is not done properly in oidc-client-js lib.

Everytime we use the id_token we first check it is not expired, if it is we try a silent sign-in, if it does not we logout the user, so he can login back again. To avoid OPF platforms issues, where the id_token has a much longer TTL then the cookie, we use the maxExpiresIn property (that was already defined for the renewal hack), to renew the token before it expires, to increase cookie TTL.


TheMaskedTurtle:
Found easier solution

PR #312, opened 2023-12-12
feat(authentication): handle id_token renewal (setInterval)

TheMaskedTurtle:


TheMaskedTurtle:
Found easier solution

PR #313, opened 2023-12-14
Sort items by subtype along with type

ne0ds:


antoinebhs (changes requested):
This PR breaks the icons in GridExplore. The icons that are on the left of the list are not displayed anymore : 
image
I'm not convinced that the approach of appending the subtype with a "/" is the best. For the case of Case and Study, it appends with undefined which might lead to some issues in the future. Did you try other approaches ?

ne0ds:
> This PR breaks the icons in GridExplore. The icons that are on the left of the list are not displayed anymore : image I'm not convinced that the approach of appending the subtype with a "/" is the best. For the case of Case and Study, it appends with undefined which might lead to some issues in the future. Did you try other approaches ?

I fixed the icons issue. For the appending of the subtybe, I don't think there could be issues when undefined is appended. Other approches would imply modifying the component in commons-ui...

antoinebhs:
* Why does "Parameters" come after Study when using alphabetical order ? 
image
* We should use a capitalize letter for "Liste d'aléas (**s**cript)"
* It's strange to use the English label for sorting, in French the sorted by type is:
image
while it should be inverted...

PR #314, opened 2023-12-19
fix(authentication): avoid silent renew on sign-in

TheMaskedTurtle:
We will have to wait the release of commons-ui with https://github.com/gridsuite/commons-ui/pull/329 to be efficient


TheMaskedTurtle:
Found easier solution

PR #316, opened 2023-12-21
Add description modification

YenguiSeddik:


src/components/dialogs/description-modification/description-modification-dialogue.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface IDescriptionModificationDialogue {
+    elementUuid: string;
+    description: string;
+    open: boolean;
+    onClose: () => void;
+}
+
+const schema = yup.object().shape({
+    [DESCRIPTION]: yup.string().nullable(),
```
AAJELLAL, line 26: The description field should not allow blank space and "" as values don't you think?

src/components/dialogs/description-modification/description-modification-dialogue.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const { reset } = methods;
+
+    const onCancel = () => {
+        reset({
+            [DESCRIPTION]: '',
+        });
+        onClose();
+    };
+
+    const onSubmit = (data: { description: string }) => {
```
AAJELLAL: It would be better if you use useCallback.

AAJELLAL (changes requested):
When I set whitespace as value a snackError is shown with the this message : 
403 Forbidden

src/components/dialogs/description-modification/description-modification-dialogue.tsx
YenguiSeddik: Description is optionnal, so i think it should be allowed

souissimai (commented):
Maybe you should truncating tooltips when the description text is too long. Please check this with PO, otherwise test ok and code ok

AAJELLAL (approved):

souissimai (approved):

souissimai (approved):

PR #318, opened 2024-01-03
Hide some contextual menu items while creating case or study

dbraquart:


src/components/toolbars/content-toolbar.js
```diff
@@ -131,22 +131,32 @@ const ContentToolbar = (props) => {
         [selectedElements, userId]
     );
 
-    const allowsDelete = useMemo(() => isUserAllowed, [isUserAllowed]);
+    const noCreationInProgress = useMemo(
+        () => selectedElements.every((el) => el.hasMetadata),
+        [selectedElements]
+    );
+
+    const allowsDelete = useMemo(
```
jamal-khey, line 139: i think allowsDelete is not particularly expensive, it use an already memoized value.

jamal-khey (approved):
Test and Code ok

antoinebhs (changes requested):
allowsCopyContingencyToScript, allowsReplaceContingencyWithScript should also be disabled, otherwise when duplicating a contingency, there can be some weird overlaps:
image

src/components/menus/content-contextual-menu.js
```diff
@@ -496,10 +497,12 @@ const ContentContextualMenu = (props) => {
...
+        return (
+            selectedElements.some(
+                (element) => element.type === ElementType.CASE
+            ) && noCreationInProgress()
         );
-    }, [selectedElements]);
+    }, [selectedElements, noCreationInProgress]);
 
     const handleDownloadCases = useCallback(async () => {
         const casesUuids = selectedElements
```
antoinebhs, line 508: instead of adding the condition in each allow[...] functions, isn't it better to add it as a parent of all the following ```if (allowsRename()){```, etc to avoid forgetting to handle this in the next addition as what happened with the Download?

dbraquart:
> allowsCopyContingencyToScript, allowsReplaceContingencyWithScript should also be disabled, otherwise when duplicating a contingency, there can be some weird overlaps: image

Only cases and studies are created async, so for them we can check and wait for metadata to be updated.
For other elements like Aleas, I dont see how to do it. I dont even know how to "overlap" actions ... maybe you can show me.

src/components/menus/content-contextual-menu.js
dbraquart: well, I dont see what we can do properly. Almost each allowXY() function is specific, so I dont see what I can put in common somewhere above.

antoinebhs (approved):

PR #320, opened 2024-01-04
Use translation for type/subtype sorting

ne0ds:


src/components/directory-content.js
```diff
@@ -673,6 +681,45 @@ const DirectoryContent = () => {
         );
     };
 
+    const sort = useCallback(
+        (datakey, isAsc, isNumeric) => {
+            const rowsCopy = [...rows].map((row, index) => [row, index]);
+            if (datakey === 'type') {
```
sBouzols: must test if datakey is null first
see code in commons-ui there is occurences of :
```sortFromProps(null, false, false);```

File: src/components/directory-content.js [hunk 2]
```diff
@@ -673,6 +681,45 @@ const DirectoryContent = () => {
         );
     };
 
+    const sort = useCallback(
+        (datakey, isAsc, isNumeric) => {
+            const rowsCopy = [...rows].map((row, index) => [row, index]);
```
sBouzols:
```suggestion
            const rowsCopy = rows.map((row, index) => [row, index]);
```
I think it's ok like that no ?

src/components/directory-content.js
```diff
@@ -673,6 +681,45 @@ const DirectoryContent = () => {
...
+                          ).localeCompare(
+                              getElementTypeTranslation(
+                                  rowIndex2[0].type,
+                                  metadata2.subtype,
+                                  metadata2.format
+                              )
+                          )
+                        : rowIndex1[0].type.localeCompare(rowIndex2[0].type);
+                });
+            } else if (isNumeric) {
```
sBouzols: Maybe add a comment here to document that we customize sort method to manage `types` and `subtypes` sorting but we need to re-implement sort for standard cases for numerical an string values.

src/components/directory-content.js
```diff
@@ -673,6 +681,45 @@ const DirectoryContent = () => {
...
+                rowsCopy.sort(
+                    (rowIndex1, rowIndex2) =>
+                        rowIndex1[0][datakey] - rowIndex2[0][datakey]
+                );
+            } else {
+                rowsCopy.sort((rowIndex1, rowIndex2) =>
+                    rowIndex1[0][datakey].localeCompare(rowIndex2[0][datakey])
```
sBouzols: are we sure to compare translated value here ? and does it work with Date formats ? Will test...

src/components/directory-content.js
```diff
@@ -639,14 +647,14 @@ const DirectoryContent = () => {
         return [...new Set(acc)];
     };
 
-    const getCurrentChildrenWithNotClickableRows = () => {
-        return currentChildren.map((child) => {
-            return {
+    const rows = useMemo(
+        () =>
+            currentChildren?.map((child) => ({
```
sBouzols, line 780: We could override the `child.type` value here in rows calling `getElementTypeTranslation` , instead of using a `typeCellRender` which is called too late after the sorting...
Then we avoid customizing the `sort` method at all...
We already assume that the data (`rows`) passed to the table is not strictly equal to `currentChildren` content (`notClickable`) then I think we could totally do this.

sBouzols (approved):
Code Review OK
Separate feature code from refactor code next time please for clarity.
Tests OK
Console Warning check OK

antoinebhs (changes requested):
Sorting works as expected but the icons corresponding to the different types (case, study, etc.) on the left of the table disappear with this PR.

antoinebhs (changes requested):
image
It looks like in this new implementation that fixes the icon on the left, the new "Description" icon disappear (see the Description icon on main branch above).

antoinebhs (approved):
Code OK
Test OK
It would be very nice to have a separate PR for the refacto next time.

sBouzols (approved):

PR #324, opened 2024-01-11
Allow the download of cases even if some selected rows are not cases.

EstherDarkish:


src/components/utils/caseUtils.ts
```diff
@@ -17,3 +27,136 @@ export const downloadCases = async (uuids: string[]) => {
...
+        if (numberOfDownloadedCases === 1) {
+            messageFirstLine = intl.formatMessage({
+                id: 'download.message.downloadedCase',
+            });
+        } else {
+            messageFirstLine = intl.formatMessage(
+                { id: 'download.message.downloadedCases' },
+                {
+                    number: numberOfDownloadedCases,
+                }
```
Mathieu-Deharbe, line 52: A bit strange to format those 2 lines differently.

src/components/utils/caseUtils.ts
```diff
@@ -17,3 +27,136 @@ export const downloadCases = async (uuids: string[]) => {
...
+        return (
+            messageFirstLine + '\n' + capitalizeFirstLetter(messageSecondLine)
+        );
+    };
+
+    const handleDownloadCases = async (selectedElements: any[]) => {
+        const casesUuids = selectedElements
+            .filter((element) => element.type === ElementType.CASE)
+            .map((element) => element.elementUuid);
+        await downloadCases(casesUuids);
```
Mathieu-Deharbe, line 150: I personnaly have nothing against async functions but Joris told me that Jon discouraged him from using it. Is there an official constraint or not?

src/components/utils/caseUtils.ts
```diff
@@ -17,3 +27,136 @@ export const downloadCases = async (uuids: string[]) => {
...
+        if (casesUuids.length !== selectedElements.length) {
+            snackInfo({
+                messageTxt: buildPartialDownloadMessage(
+                    casesUuids.length,
+                    selectedElements
+                ),
+            });
+        }
```
Mathieu-Deharbe, line 158: So the snackbar is only displayed when there were non situations selected ? Too bad : it looks much nicer than the navigator notification.

src/translations/en.json
```diff
@@ -257,8 +257,17 @@
   "country": "Country",
   "voltageRegulatorOn": "Voltage regulation",
   "PlannedActivePowerSetPoint": "Planning active power set point",
-  "downloadCases": "Download Case(s)",
-
+  "download.button": "Download",
```
Mathieu-Deharbe, line 260: Why did you add a subcategory "button" here ? Seems to me that such a generic word could be reused anywhere.

Same remark for "download.message.filters": "filters",
  "download.message.lists": "lists",
  "download.message.studies": "studies",

Mathieu-Deharbe (approved):

src/components/utils/caseUtils.ts
EstherDarkish: I reused code from @ne0ds and checked beforehand with him.
Maybe we should do some refactoring, but not in this pull request.

src/translations/en.json
EstherDarkish: Yes, there should be work done with the application's translation keys, they are over the place. Because I work on another ticket modifying them, I'd like to structure them better, and we have to start somewhere (so, let's start here).

src/components/utils/caseUtils.ts
EstherDarkish: Yes

src/components/utils/caseUtils.ts
EstherDarkish: Yes, ideally we should spell out numbers under 10, but that would be too much code for this anectdotical message

src/components/utils/caseUtils.ts
```diff
@@ -1,4 +1,14 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
 import { downloadCase, getCaseOriginalName } from '../../utils/rest-api';
+import { ElementType } from '../../utils/elementType';
+import { useIntl } from 'react-intl';
+import { useSnackMessage } from '@gridsuite/commons-ui';
 
 export const downloadCases = async (uuids: string[]) => {
```
sBouzols: If it's used only locally now (because you export useDownloadUtils now) maybe remove this export here

EstherDarkish: Yes, done

sBouzols (approved):
We could do better for plural management and multiple values management for translation but ok for now before defining Translation guidelines.
Code Review OK

PR #325, opened 2024-01-12
Fix description field

YenguiSeddik:


src/translations/en.json
```diff
@@ -242,5 +242,5 @@
   "downloadCases": "Download Case(s)",
 
   "descriptionModificationError": "An error when modifying the description",
-  "descriptionModificationDialog": "Description modification"
```
AbdelHedhili, line 269: this wasn't used ?

AAJELLAL (changes requested):
- The application crashes when creating a study or situation regardless of adding the description or not.
- The description counter needs some margin at the bottom.

AAJELLAL (approved):
Tests & code OK.

AbdelHedhili (approved):

src/translations/en.json
YenguiSeddik: No, it was used in the description modification dialog but was replaced by "description": "Description",

souissimai (approved):
test ok

PR #327, opened 2024-01-16
Fixes the contextual selection of elements.

EstherDarkish:
Fixes the contextual selection of elements.
NOTE : Zimbra works like Google and others and as such, contextualMixPolicies.ZIMBRA is removed.


src/components/directory-content.js
```diff
@@ -719,60 +753,37 @@ const DirectoryContent = () => {
...
-                ctxtUuid === null ||
-                selectedUuids.has(ctxtUuid)
-            ) {
-                acc = acc.concat(
+    const getSelectedChildren = () => {
+        let selectedChildren = [];
+        if (currentChildren?.length > 0) {
+            // Adds the previously selected elements
+            if (selectedUuids?.size) {
+                selectedChildren = selectedChildren.concat(
```
sBouzols: With what do you concat ? an empty array ? maybe this form is enought :
```
let selectedChildren = currentChildren.filter(...).map(...)
```

src/components/directory-content.js
```diff
@@ -719,60 +753,37 @@ const DirectoryContent = () => {
...
-                            return Object.assign(
-                                {
-                                    subtype:
-                                        childrenMetadata[child2.elementUuid],
-                                },
-                                child2
-                            );
+                        .map((child) => {
+                            return {
+                                subtype: childrenMetadata[child.elementUuid],
```
sBouzols: Where `.subtype` ? seems like an existing bug to me ... We won't have a lot of subtypes in ContentContextualMenu or in ContentToolbar I think...

src/components/directory-content.js
```diff
@@ -719,60 +753,37 @@ const DirectoryContent = () => {
...
-                mayChange &&
-                contextualMixPolicy === contextualMixPolicies.BIG
-            ) {
-                setSelectedUuids(null);
+            }
+
+            // Adds the active element
+            if (activeElement) {
+                selectedChildren.push({
+                    subtype:
```
sBouzols, line 793: I prefer : 
```
{
    ...activeElement,
    subtype: childrenMetadata[activeElement.elementUuid]?.subtype,
}
```
because it clearly says that we overwrite any existing or not `subtype` value in `activeElement`, otherwise you could think reading the code that `subtype` has the value of `childrenMetadata[activeElement.elementUuid]?.subtype` but if this key is present in the `activeElement` object then it will overwrite and prevail.

src/components/menus/content-contextual-menu.js
```diff
@@ -586,6 +586,7 @@ const ContentContextualMenu = (props) => {
                 icon: <FileCopyIcon fontSize="small" />,
             });
         }
+
```
sBouzols, line 589: useless line break which mark this component as modified by this commit but it's not.

EstherDarkish: That's the most important part of this PR in fact, because in this code block, that was the only place without a line break :)
Other line breaks like this missing one are on lines 518, 527, 538, 548, 565, 576, 600 and 612.

src/components/directory-content.js
EstherDarkish: Yes, nice catch

src/components/directory-content.js
EstherDarkish: Ok, added the missing `?.subtype`

src/components/directory-content.js
EstherDarkish: Done

src/components/directory-content.js
```diff
@@ -295,6 +304,24 @@ const DirectoryContent = () => {
...
+                        if (
+                            selectedUuids?.size &&
+                            element?.elementUuid &&
+                            !selectedUuids.has(element.elementUuid)
+                        ) {
+                            setSelectedUuids(new Set());
+                        }
+                    } else {
+                        // If some elements were already selected, we add the active element to the selected list if not already in it.
+                        if (selectedUuids?.size && element?.elementUuid) {
```
Mathieu-Deharbe: Shouldn't you add `&& !selectedUuids.has(element.elementUuid)` ? It seems to me that in this case it is better to not create a Set (but I understand that it works anyway).

src/components/directory-content.js
```diff
@@ -282,6 +282,15 @@ const DirectoryContent = () => {
...
+            BIG: 'GoogleMicrosoft', // if !selectedUuids.has(selected.Uuid) deselects selectedUuids
+            ALL: 'All', // union of activeElement.Uuid and selectedUuids (currently implemented)
+        }),
+        []
+    );
+    const contextualMixPolicy = contextualMixPolicies.ALL;
+
     const onContextMenu = useCallback(
         (event) => {
             const element = currentChildren.find(
```
Mathieu-Deharbe, line 296: Could be named `hoveredElt`, I think it would to be more precise/readable.

Mathieu-Deharbe (approved):

src/components/directory-content.js
EstherDarkish: I'd rather not change too much outside of scope to keep the PR simple

src/components/directory-content.js
EstherDarkish: Yes, nice catch.
Note : the updated version could easily have the policy test inside the element/selected test, but I'd rather keep the policy test first, for clarity.

sBouzols (approved):
Code Review OK

PR #332, opened 2024-01-22
Add voltage level and substation support for expert filters.

flomillot:
Add low voltage limit and high voltage limit fields in expert filter type and translations.


src/components/dialogs/filter/expert/expert-filter-constants.ts
```diff
@@ -227,4 +247,17 @@ export const fields: Record<string, Field[]> = {
...
     LOAD: [FIELDS_OPTIONS.ID],
+    VOLTAGE_LEVEL: [
+        FIELDS_OPTIONS.ID,
+        FIELDS_OPTIONS.NAME,
+        FIELDS_OPTIONS.NOMINAL_VOLTAGE,
+        FIELDS_OPTIONS.LOW_VOLTAGE_LIMIT,
+        FIELDS_OPTIONS.HIGH_VOLTAGE_LIMIT,
+        FIELDS_OPTIONS.COUNTRY,
+    ],
+    SUBSTATIONS: [
```
Meklo:
```suggestion
    SUBSTATION: [
```
typo ? Otherwise fields don't show up when selecting substations

src/components/dialogs/filter/expert/expert-filter-form.tsx
```diff
@@ -76,7 +81,12 @@ export const expertFilterSchema = {
 };
 
 function isSupportedEquipmentType(equipmentType: string): boolean {
-    return equipmentType === Generator.type || equipmentType === Load.type;
+    return (
+        equipmentType === Generator.type ||
+        equipmentType === Load.type ||
+        equipmentType === VoltageLevel.type ||
+        equipmentType === Substation.type
+    );
```
Meklo:
```suggestion
        return Object.values(EXPERT_FILTER_EQUIPMENTS)
        .map((equipments) => equipments.id)
        .includes(equipmentType);
```
how about future proofing it ?

src/components/dialogs/filter/expert/expert-filter-constants.ts
flomillot: Ah yes my bad thanks

src/components/dialogs/filter/expert/expert-filter-form.tsx
flomillot: You're right I change it

Meklo (approved):

lilian-houdelet (approved):
Code OK
Test OK

PR #334, opened 2024-01-24
Add description to contingency list and filter création

YenguiSeddik:


src/utils/rest-api.js
```diff
@@ -571,12 +571,13 @@ export function rootDirectoryExists(directoryName) {
...
     contingencyListType,
     contingencyListName,
+    description,
     formContent,
     parentDirectoryUuid
 ) {
     console.info('Creating a new contingency list...');
     let urlSearchParams = new URLSearchParams();
-    urlSearchParams.append('description', ''); // TODO Remove this when the backend does not need it anymore
+    urlSearchParams.append('description', description); // TODO Remove this when the backend does not need it anymore
```
souissimai: Maybe you should remove comment ( it does not make sense ) since we should send description value to the backend

src/components/directory-content.js
```diff
@@ -583,9 +583,7 @@ const DirectoryContent = () => {
             );
             return (
                 <>
-                    {isElementCaseOrStudy(element.type) && (
-                        <Box sx={styles.cell}>{icon}</Box>
-                    )}
+                    <Box sx={styles.cell}>{icon}</Box>
```
souissimai: exclude voltage init params

AAJELLAL (approved):
test ok & code ok.

souissimai (approved):
code and test ok

PR #339, opened 2024-01-29
Add Custom filter for lines

lilian-houdelet:


flomillot (changes requested):
Countries does not work
https://github.com/gridsuite/gridexplore-app/blob/b38ba7e5225bdfb6151ff512e2505b2d5b903d5a/src/components/utils/rqb-inputs/value-editor.tsx#L24

src/components/utils/rqb-inputs/value-editor.tsx
```diff
@@ -21,7 +21,13 @@ const ValueEditor = (props: ValueEditorProps) => {
         // No value needed for this operator
         return null;
     }
-    if (props.field === FieldType.COUNTRY) {
+    if (
```
flomillot, line 24: if (
        [FieldType.COUNTRY, FieldType.COUNTRY_1, FieldType.COUNTRY_2].includes(
            props.field as FieldType
        )
    )

Better than call toString() everytime, no ?

src/components/dialogs/filter/expert/expert-filter-utils.ts
```diff
@@ -93,19 +101,36 @@ export const getOperators = (fieldName: string, intl: IntlShape) => {
     return defaultOperators;
 };
 
+function changeValueUnit(value: any, field: FieldType) {
+    if (microUnits.includes(field as FieldType)) {
```
flomillot:
```suggestion
    if (microUnits.includes(field)) {
```

src/components/dialogs/filter/expert/expert-filter-utils.ts
```diff
@@ -93,19 +101,36 @@ export const getOperators = (fieldName: string, intl: IntlShape) => {
...
 ): RuleGroupTypeExport {
     function transformRule(rule: CustomRuleType): RuleTypeExport {
         const isValueAnArray = Array.isArray(rule.value);
+
+        let updatedValue = rule.value;
+
+        if (rule.operator !== OperatorType.EXISTS) {
+            updatedValue = changeValueUnit(rule.value, rule.field as FieldType);
+        }
+
```
flomillot: Why this code ?

src/components/dialogs/filter/expert/expert-filter-utils.ts
```diff
@@ -93,19 +101,36 @@ export const getOperators = (fieldName: string, intl: IntlShape) => {
...
         return {
             field: rule.field as FieldType,
             operator: Object.values(OPERATOR_OPTIONS).find(
                 (operator) => operator.name === rule.operator
             )?.customName as OperatorType,
             value:
                 !isValueAnArray && rule.operator !== OperatorType.EXISTS
-                    ? rule.value
+                    ? updatedValue
```
flomillot:
```suggestion
                    ? changeValueUnit(rule.value, rule.field as FieldType)
```
Why not just this ?

src/components/dialogs/filter/expert/expert-filter-utils.ts
```diff
@@ -93,19 +101,36 @@ export const getOperators = (fieldName: string, intl: IntlShape) => {
...
                     : undefined,
             values: isValueAnArray ? rule.value : undefined,
```
flomillot: What about `values` ?

src/components/dialogs/filter/expert/expert-filter-utils.ts
```diff
@@ -140,13 +165,19 @@ export function importExpertRules(
             // values is a Set on server side, so need to sort
             if (rule.dataType === DataType.NUMBER) {
                 return rule.values
-                    .map((value) => parseFloat(value as string))
+                    .map((value) => {
```
flomillot: You can keep the previous code and add a new map right after, then we avoid duplicate code.

flomillot (changes requested):

src/components/dialogs/filter/expert/expert-filter-utils.ts
lilian-houdelet: we don't have any values for the operator EXISTS

lilian-houdelet: Not mandatory with the changes

src/components/dialogs/filter/expert/expert-filter-utils.ts
lilian-houdelet: I forgot, I'll update it

flomillot (approved):

PR #343, opened 2024-02-01
Add filters free properties support

Meklo:


src/components/dialogs/filter/criteria-based/filter-properties.tsx
```diff
@@ -176,33 +145,16 @@ function FilterProperties() {
                     <FormattedMessage id={'FreePropsCrit'}>
                         {(txt) => <h3>{txt}</h3>}
                     </FormattedMessage>
```
achour94, line 180: should close your grid item here, else you have you button centered

src/components/dialogs/filter/criteria-based/filter-properties.tsx
```diff
@@ -176,33 +145,16 @@ function FilterProperties() {
...
                     )}
-                </Grid>
-                {filterProperties.map((prop, index) => (
-                    <ListItem key={prop.id}>
-                        <FilterProperty
-                            index={index}
-                            valuesFields={valuesFields}
-                            predefined={fieldProps}
-                            handleDelete={remove}
+                    {!isForSubstation && (
```
achour94, line 189: the substation free properties should be visible only for Loads

src/components/dialogs/filter/criteria-based/filter-properties.tsx
```diff
@@ -100,74 +96,47 @@ export const filterPropertiesYupSchema = {
...
+                }
+                const names = properties! // never null / undefined
+                    .filter((prop) => !!prop[PROPERTY_NAME])
+                    .map((prop) => prop[PROPERTY_NAME]);
+                return areArrayElementsUnique(names);
+            }
+        ),
 };
 
-function fetchPredefinedProperties() {
```
achour94, line 105: this function is called for each child component, in my opinion it will be better to let it here so we do the fetch once and we pass the predefined properties as a prop to the free properties sub-components

src/components/dialogs/filter/criteria-based/filter-free-properties.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+        return studyMetadata?.predefinedEquipmentProperties?.substation;
+    });
+}
+
+interface FilterFreePropertiesProps {
+    freePropertiesType: string;
+}
+
+function FilterSubstationProperties({
```
achour94: this component is used as a generic , so you should remove substation from the component name

src/components/dialogs/filter/criteria-based/filter-free-properties.tsx
```diff
@@ -0,0 +1,144 @@
...
+                'Study entry could not be found in metadatas'
+            );
+        }
+
+        return studyMetadata?.predefinedEquipmentProperties?.substation;
+    });
+}
+
+interface FilterFreePropertiesProps {
+    freePropertiesType: string;
```
achour94: this component is designed to handle two types of freeProperties (FREE_FILTER_PROPERTIES and SUBSTATION_FILTER_PROPERTIES) not any string

achour94 (changes requested):

src/components/dialogs/filter/criteria-based/filter-properties.tsx
Meklo: The issue is that closing the Grid tag here creates a huge gap between the section title and subsections titles
I've added an xs attribute to the button, tell me if that looks alright to you

src/components/dialogs/filter/criteria-based/filter-properties.tsx
Meklo: I've made a change so that substations filtering are now handled by the attribute freeProperties instead of substationFreeProperties, otherwise it's redundant and conceptually wrong, we can further talk about it if you want

src/components/dialogs/filter/criteria-based/filter-free-properties.tsx
Meklo: My bad, forgot to remove it, done

src/components/dialogs/filter/criteria-based/filter-free-properties.tsx
Meklo: I've replaced it by an enum

src/components/dialogs/filter/criteria-based/filter-properties.tsx
Meklo: Agreed, I've mutualized it in a state in FilterProperties

achour94 (approved):

YenguiSeddik (approved):

PR #344, opened 2024-02-06
add MODIFICATION type

dbraquart:


FranckLecuyer (changes requested):
Tests:
The generated modification's name and the generated modification's description are fixed and not translated between french/english : it seems that they are in the language that was used in GridStudy when they were generated ...

The order of the modifications added to a node when imported in GridStudy from GridExplore seems to be random ... it is not the display order of the modifications in the directory (maybe the selection order of those modifications ?)

dbraquart:
> Tests: The generated modification's name and the generated modification's description are fixed and not translated between french/english : it seems that they are in the language that was used in GridStudy when they were generated ...
> 
> The order of the modifications added to a node when imported in GridStudy from GridExplore seems to be random ... it is not the display order of the modifications in the directory (maybe the selection order of those modifications ?)

point 1) elements names and descriptions are not translatable. I just use the current language when I create the elements. I think this is the expected behaviour, but I will circle-back with the PO.
point 2) yes I'm not sure what is the order we use, but it is explicitly out of scope for now

FranckLecuyer (approved):

PR #345, opened 2024-02-06
send case format in the request for case creation

jamal-khey:


src/utils/rest-api.js
```diff
@@ -442,6 +444,7 @@ export function duplicateStudy(
     studyDescription,
     sourceStudyUuid,
     parentDirectoryUuid
+    //(jamal) add case format here also ?
```
flomillot: Forgot ?

jamal-khey: yes i fixed it

src/components/dialogs/create-study-dialog/create-study-dialog.js
```diff
@@ -145,9 +146,11 @@ const CreateStudyDialog = ({ open, onClose, providedExistingCase }) => {
                     setValue(FORMATTED_CASE_PARAMETERS, formattedParams, {
                         shouldDirty: true,
                     });
+                    setCaseFormat(result.formatName);
```
anistouri: Why don't you add it with the other fields?
```suggestion
                    setValue(FORMAT_NAME, result.formatName);
```

anistouri (approved):

src/components/dialogs/create-study-dialog/create-study-dialog.js
jamal-khey: done, i removed the state

src/components/dialogs/create-study-dialog/create-study-dialog-utils.js
```diff
@@ -34,4 +36,5 @@ export const createStudyDialogFormValidationSchema = yup.object().shape({
     [CASE_UUID]: yup.string().required(),
     [CASE_FILE]: yup.mixed().nullable().required(),
     [DIRECTORY]: yup.string().required(),
+    [CASE_FORMAT]: yup.string().optional(),
```
flomillot: I don't think we use optional on string in our projects, why adding it as we use empty string ?
https://github.com/jquense/yup#schemaoptional-schema

anistouri: Why do you want the field to be optional? Are there cases where we can be without the format? I believe it should be required

jamal-khey: i changed it to be required

flomillot (approved):
Code OK

anistouri (approved):

Meklo (approved):

Meklo:
I may have approved a bit hastily, I noticed afterward that setting CASE_FORMAT to required has produced a side effect.
Now when importing a valid case the form doesn't automatically enable the Validate button, do you get that too ?

jamal-khey:
> I may have approved a bit hastily, I noticed afterward that setting CASE_FORMAT to required has produced a side effect. Now when importing a valid case the form doesn't automatically enable the Validate button, do you get that too ?

I wasn't aware of this side effect. but i think it does not affect the user experience that much, the button become quickly available

PR #346, opened 2024-02-07
add snackbar on maintenance message

AbdelHedhili:


jamal-khey (approved):
test and code Ok

flomillot:
In meeting last time we said we should create a new hook inside commons-ui instead of duplicating the code in both gridstudy and gridexplore.
We already have something similar that we could ajust or get inspiration from, and it's something we need in every front app.
https://github.com/gridsuite/commons-ui/blob/520b575880b7bb5fbe582d34a0f4b924b9274573/src/hooks/useSnackMessage.js#L12

AbdelHedhili:
> In meeting last time we said we should create a new hook inside commons-ui instead of duplicating the code in both gridstudy and gridexplore. We already have something similar that we could ajust or get inspiration from, and it's something we need in every front app. https://github.com/gridsuite/commons-ui/blob/520b575880b7bb5fbe582d34a0f4b924b9274573/src/hooks/useSnackMessage.js#L12

agreed, it was just a fast first version. We can do the refacto in another or in this one if you prefer to use commons-ui

PR #351, opened 2024-02-09
Remove the arrows in the expert filter's numerical fields

EstherDarkish:


antoinebhs (approved):
Code OK and consistent with https://github.com/gridsuite/gridstudy-app/blob/ec2d7b48fc58471b62a4f352eff5ba813bfa90da/src/components/spreadsheet/utils/equipment-table-editors.js#L263

src/components/utils/rqb-inputs/value-editor.tsx
```diff
@@ -38,10 +51,12 @@ const ValueEditor = (props: ValueEditorProps) => {
         return <TextValueEditor {...props} />;
     }
     return (
-        <MaterialValueEditor
-            {...props}
-            title={undefined} // disable the tooltip
-        />
+        <Box sx={props.inputType === 'number' ? styles.noArrows : undefined}>
```
ne0ds, line 54: I think that we don't need to check for input type and that we can apply the style in any case.

EstherDarkish: I'd rather not add unecessary styling that could be confusing down the line.

PR #359, opened 2024-02-16
Make case_format optional in createStudyDialogFormValidationSchema

jamal-khey:
Update createStudyDialogFormValidationSchema to make the `case_format` field optional instead of required. This change allows users to validate the form without being blocked by the `case_format` field.


dbraquart (approved):
code:
tests: ok

PR #371, opened 2024-03-04
Restore elements in active directory instead of selected directory

YenguiSeddik:


src/components/dialogs/stashed-elements/stashed-elements-dialog.tsx
```diff
@@ -27,6 +27,7 @@ interface IStashedElementsDialog {
     onClose: () => void;
     stashedElements: any[];
     onStashedElementChange: () => any[];
+    activeDirectory?: any;
```
Meklo:
```suggestion
    activeDirectory?: IDirectory;
```
This would fit I guess

YenguiSeddik: done

Meklo (approved):

src/components/dialogs/stashed-elements/stashed-elements-dialog.tsx
```diff
@@ -27,6 +27,7 @@ interface IStashedElementsDialog {
     onClose: () => void;
     stashedElements: any[];
     onStashedElementChange: () => any[];
+    activeDirectory?: IDirectory;
```
achour94: The StashedElementsDialog component should not be responsible for determining the directory where the elements will be restored. It would be cleaner and more robust to pass the directory as a prop from the parent component, allowing the parent to specify which directory (selected or active) should be used.

YenguiSeddik: done

achour94 (approved):

PR #382, opened 2024-03-18
User can select multiple elements with ctrl or shift clicking

klesaulnier:


EstherDarkish (approved):
Test OK code OK

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
+import { useCallback, useEffect, useState } from 'react';
```
sBouzols, line 7: license header

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Hook to deal with list of checkboxes
```
sBouzols: not only checkboxes, for example here it manage multi-selection in a table when clicking on the line as well.

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+ * @param elementIds list of all ids used for selection, whether they are selected or not
+ * for "handleShiftAndCtrlClick" to work, this list needs to be sorted in the same order as it is displayed
+ */
+export const useMultiselect = (elementIds: string[]) => {
```
sBouzols, line 14: I would rename `elementIds` to something like `options` for similar APIs in MUI for example.
We could accept to manage only string[] of ids for now, but it could become more generic accepting Object[] as well with another `comparator` prop to manage equality between objects. or even UUID[]

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+    const [selectedIds, setSelectedIds] = useState<Set<string>>(new Set());
+    // used for shift clicking selection, stores last clicked element for selection
+    const [lastSelectedElementId, setLastSelectedElementId] =
+        useState<string>();
+
+    useEffect(() => {
+        setSelectedIds(new Set());
```
sBouzols: I know it's like that on the previous code, but if we add an item in the list (by notification for example) we completly lost the selection. Could be better, managing intersection between `elementIds` and `selectedIds`

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+        setLastSelectedElementId(undefined);
```
sBouzols: I would prefer setting to `null`

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
+import { useCallback, useEffect, useState } from 'react';
+
+/**
+ * Hook to deal with list of checkboxes
+ * @param elementIds list of all ids used for selection, whether they are selected or not
+ * for "handleShiftAndCtrlClick" to work, this list needs to be sorted in the same order as it is displayed
+ */
+export const useMultiselect = (elementIds: string[]) => {
+    const [selectedIds, setSelectedIds] = useState<Set<string>>(new Set());
```
sBouzols: I think it's better to manage `Set` where it's optimized to have it, checking existence (has(), unicity etc...) but to return a simple ordered array. Then this state should be a an array for me.

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+     * @param elementId id of element to toggle
+     * @param forceState if defined, it will force element state instead of toggling (false -> unselect element ; true -> select element)
+     */
+    const toggleSelection = useCallback(
+        (elementId: string, forceState?: boolean) => {
+            let element = elementIds?.find((id) => id === elementId);
+            if (element === undefined) {
+                return;
+            }
+            let newSelection = new Set(selectedIds);
```
sBouzols: then here you manage a local Set from the array state `selectedIds` .

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+                    newSelection.add(elementId);
+                }
+            } else {
+                if (forceState) {
+                    newSelection.add(elementId);
+                } else {
+                    newSelection.delete(elementId);
+                }
+            }
+            setSelectedIds(newSelection);
```
sBouzols: and here you set the array state `setSelectedIds([...newSelection]);

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+
+            elementToSelectIds.forEach((elementToSelectId) => {
+                let element = elementIds?.find(
+                    (id) => id === elementToSelectId
+                );
+                if (element !== undefined) {
+                    newSelection.add(elementToSelectId);
+                }
+            });
+            setSelectedIds(newSelection);
```
sBouzols:
```suggestion
            const intersection = elementToSelectIds.filter(value => elementIds.includes(value)); // check consistency
            setSelectedIds([...new Set([...selectedIds, ...intersection)]; // union for selection
```

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+        },
+        [selectedIds]
+    );
+
+    const handleShiftClick = useCallback(
+        (clickedElementId: string) => {
+            // remove text selection due to shift clicking
+            window.getSelection()?.empty();
```
sBouzols, line 85: I would move this, because it's not linked with the multiselection computation but with the render of some components associated. Maybe move it in the event handler on Click of the `Box` (the line in the table).

src/utils/use-multiselect.ts
sBouzols: Not even necessary to manage here the consistency, because it's a private function inside this hook which is called only by `handleShiftClick`. Maybe manage consistency between `elementIds` and entries only at one place.

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+        setLastSelectedElementId(undefined);
+    }, [elementIds]);
+
+    /**
+     * toggle selection for one element
+     * @param elementId id of element to toggle
+     * @param forceState if defined, it will force element state instead of toggling (false -> unselect element ; true -> select element)
+     */
+    const toggleSelection = useCallback(
+        (elementId: string, forceState?: boolean) => {
```
sBouzols: the forceState parameter is never used

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+        clickedElementId: string
+    ) => {
+        if (clickEvent.shiftKey) {
+            // if row is clicked while shift is pressed, range of rows selection is toggled, depending on clicked element state
+            handleShiftClick(clickedElementId);
+            // nothing else happens, hence the return
+            return;
+        }
+
+        if (clickEvent.ctrlKey) {
```
sBouzols, line 139: Ctrl+click or click should toggle selection I think. this custom hook should not know that there is something special when clicking on a row or a checkbox.

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+            toggleSelection,
+        ]
+    );
+
+    const handleShiftAndCtrlClick = (
+        clickEvent: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+        clickedElementId: string
+    ) => {
+        if (clickEvent.shiftKey) {
+            // if row is clicked while shift is pressed, range of rows selection is toggled, depending on clicked element state
```
sBouzols: this custom hook is not relative to any components then don't mention row or anything I think.

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+        }
+    };
+
+    /**
+     * toggle selection for all element depending on current selectedIds
+     * @param elementsToSelectIds if defined, it will toggle only elementsToSelectIds instead of all elementIds
+     * @param forceSelectedIds if true, it will set selection to elementsToSelectIds/elementIds without checking current selectedIds
+     */
+    function toggleSelectAll(
+        elementsToSelectIds?: string[],
```
sBouzols: toggleSelectAll should simply do something like 
```
setSelectedIds(options)
// or
setSelectedIds([])
```
not parameters or it becomes `toggleSelectAllOrPartialItDependsOfTheParametersYouPass`

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+        if (selectedIds.size === 0 || forceSelectedIds) {
+            setSelectedIds(new Set(elementsToSelectIds ?? elementIds));
+        } else {
+            setSelectedIds(new Set());
+        }
+    }
+
+    return [
+        selectedIds,
+        toggleSelection,
```
sBouzols, line 162: I don't think we need to return this, we already return `handleShiftAndCtrlClick`, don't need more I think.

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,162 @@
...
+        toggleSelectAll,
+        handleShiftAndCtrlClick,
```
sBouzols, line 165: rename to `handleSelectionEvent` only. Because this custom hook implement ctrl and shift logic, and it's not the matter of the user.

src/components/directory-content.js
```diff
@@ -401,72 +414,91 @@ const DirectoryContent = () => {
         [appsAndUrls]
     );
 
+    const handleClickElementCheckbox = useCallback(
```
sBouzols, line 415: to remove

src/components/directory-content.js
```diff
@@ -750,10 +758,9 @@ const DirectoryContent = () => {
         const elementUuid = cellData.rowData['elementUuid'];
         return (
             <Box
-                onClick={(e) => {
-                    toggleSelection(elementUuid);
-                    e.stopPropagation();
-                }}
+                onClick={(clickEvent) =>
+                    handleClickElementCheckbox(clickEvent, elementUuid)
```
sBouzols, line 734:
```suggestion
                    {
                        handleShiftAndCtrlClick(clickEvent, elementUuid);
                        clickEvent.stopPropagation();
                    }
                    
```

src/components/directory-content.js
```diff
@@ -698,39 +730,15 @@ const DirectoryContent = () => {
...
     function selectionHeaderRenderer() {
         return (
             <Box
                 onClick={(e) => {
-                    toggleSelectAll();
+                    toggleSelectAll(
+                        currentChildren
+                            .filter((e) => !e.uploading)
+                            .map((c) => c.elementUuid)
+                    );
```
sBouzols:
```suggestion
                    toggleSelectAll();
```

src/components/directory-content.js
```diff
@@ -321,7 +333,7 @@ const DirectoryContent = () => {
                             element?.elementUuid &&
                             !selectedUuids.has(element.elementUuid)
                         ) {
-                            setSelectedUuids(new Set());
+                            toggleSelectAll(new Set(), true);
```
sBouzols: arf... maybe expose a `clearSelection` function from the custom hook ?

src/components/directory-content.js
```diff
@@ -332,7 +344,7 @@ const DirectoryContent = () => {
                         ) {
                             let updatedSelectedUuids = new Set(selectedUuids);
                             updatedSelectedUuids.add(element.elementUuid);
-                            setSelectedUuids(updatedSelectedUuids);
+                            toggleSelectAll(updatedSelectedUuids, true);
```
sBouzols: arf... maybe expose `selectElements` from the custom hook but manage `Set` and all inside the hook
```
    selectElements([element.elementUuid]);
```

src/components/directory-content.js
```diff
@@ -332,7 +344,7 @@ const DirectoryContent = () => {
                         ) {
```
sBouzols, line 343: it is the exact same if statement... maybe reorder the code to have something like tha : 
```
if (
    selectedUuids?.size &&
    element?.elementUuid &&
    !selectedUuids.has(element.elementUuid)
) {
    if (contextualMixPolicy === contextualMixPolicies.BIG) { // never used
        clearSelection();
    }
    selectElements([element.elementUuid]);
}
```

src/components/directory-content.js
```diff
@@ -160,12 +161,23 @@ const DirectoryContent = () => {
...
 
-    const [selectedUuids, setSelectedUuids] = useState(new Set());
-
     const currentChildren = useSelector((state) => state.currentChildren);
     const currentChildrenRef = useRef();
     currentChildrenRef.current = currentChildren;
 
+    const currentChildrenUuids = useMemo(
+        () =>
+            currentChildren ? currentChildren.map((e) => e.elementUuid) : [],
```
sBouzols: maybe `.filter((e) => !e.uploading)` here to avoid selecting uploading elements.
With this functionality, I think we should disable or add an overlay over the uploading line because the User should know...

sBouzols (commented):
Some code remarks
I know some are refactoring stuff but for better code quality I think.
Not tested... Will

src/utils/use-multiselect.ts
klesaulnier: done

src/utils/use-multiselect.ts
klesaulnier: what description would you use without making the hook description difficult to understand ?
it's still a list of checkbox, even if we can check/uncheck by clicking somewhere else

src/utils/use-multiselect.ts
klesaulnier: if I don't use the word "checkbox" nor "row", what do you suggest ?

src/utils/use-multiselect.ts
klesaulnier: for now, this hooks work well with ids and is supposed to work this way, if another component need it to work with something else than ID, we'll change it accordingly

src/utils/use-multiselect.ts
klesaulnier: done

src/utils/use-multiselect.ts
klesaulnier: It's mostly kept from old code
I'll try without checking it

src/utils/use-multiselect.ts
klesaulnier: I agree, fixed. It was the previous behaviour, I might need to go back to this behaviour if filtering uploading lines doesn't work straight away

src/components/directory-content.js
klesaulnier: not related to my PR

src/components/directory-content.js
klesaulnier: Testing to filter right here, but it was the previous behaviour, we're taking risks to create bugs

src/components/directory-content.js
klesaulnier: done

src/components/directory-content.js
klesaulnier: done

src/utils/use-multiselect.ts
klesaulnier: for now I'll leave it like this, it's not defined by the PO

src/utils/use-multiselect.ts
klesaulnier: done

src/utils/use-multiselect.ts
klesaulnier: done

src/utils/use-multiselect.ts
klesaulnier: done

src/utils/use-multiselect.ts
klesaulnier: removed

src/utils/use-multiselect.ts
klesaulnier: I don't understand your point here ? 
This method is use for ctrl and shift click only

src/utils/use-multiselect.ts
klesaulnier: handleShiftAndCtrlClick is only to handle shift and ctrl clicks, it's not the same

src/utils/use-multiselect.ts
klesaulnier: I want this method to only handle shift and ctrl clicks, renaming it would make it unclear

src/components/directory-content.js
klesaulnier: I don't think I should, I want this method to work this way

src/components/directory-content.js
klesaulnier: done

src/utils/use-multiselect.ts
klesaulnier: I agree in some way, but that would mean calling "window.getSelection()?.empty();" each time you use handleShiftClick somewhere. Adding an event handler in the middle of the component everytime you want to enable shift clicking does not sound better

src/components/directory-content.js
EstherDarkish: For context : this was done this way to have the "BIG" or other policy in the first level of `if` clauses, to be easier to read and understand.

EstherDarkish (approved):

src/utils/use-multiselect.ts
sBouzols: component ?

src/utils/use-multiselect.ts
sBouzols:
```suggestion
 * Hook to deal with list of components multiselection
```

src/utils/use-multiselect.ts
sBouzols: ok we could keep it here for now.

src/utils/use-multiselect.ts
sBouzols: by the way I figured out we need it to manage selection when handling right click on a component... but it could be even easier if this hook manage rigth click event as well :P ... ok ok later...

src/utils/use-multiselect.ts
sBouzols: I simply want to point out that the hook could manage all kinds of selection events. simple selection (click), Ctrl selection, shift selection, right click selection, later a specific rule for ctrl+shift+alt selection I don't now... but all this logic could be encapsulated in the hook.

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,164 @@
...
+     * @param forceState if defined, it will force element state instead of toggling (false -> unselect element ; true -> select element)
+     */
+    const toggleSelection = useCallback(
+        (elementToToggleId: string) => {
+            let element = elementIds?.find((id) => id === elementToToggleId);
+            if (element === undefined) {
+                return;
+            }
+
+            selectedIds.indexOf(elementToToggleId);
```
sBouzols: to remove (copy/paste problem)

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,164 @@
...
+        (elementsToSelectIds: string[]) => {
+            elementsToSelectIds
+                .filter(
+                    (elementToSelectId) =>
+                        !selectedIds.includes(elementToSelectId)
+                )
+                .forEach((elementToSelectId) => {
+                    selectedIds.push(elementToSelectId);
+                });
+            setSelectedIds([...selectedIds]);
```
sBouzols, line 67:
```suggestion
            setSelectedIds([...new Set([...selectedIds, ...elementsToSelectIds])])
```

src/utils/use-multiselect.ts
```diff
@@ -0,0 +1,164 @@
...
+        [selectedIds]
+    );
+
+    const removeElementsFromSelection = useCallback(
+        (elementsToUnselectIds: string[]) =>
+            setSelectedIds([
+                ...selectedIds.filter(
+                    (id) => !elementsToUnselectIds.includes(id)
+                ),
+            ]),
```
sBouzols, line 78:
```suggestion
            const elementsToUnselectIdsSet = new Set(elementsToUnselectIds);
            setSelectedIds([
                ...selectedIds.filter(
                    (id) => !elementsToUnselectIdsSet.has(id)
                ),
            ]),
```
for performance

src/utils/use-multiselect.ts
sBouzols: deal

src/utils/use-multiselect.ts
sBouzols: it could become a more global selectionEventHandler which manage more event types.

src/components/directory-content.js
sBouzols: later

src/components/directory-content.js
sBouzols: later

src/utils/use-multiselect.ts
sBouzols: We cannot use Set here because we need to keep order to use lastSelectedElementId ...
My bad

src/utils/use-multiselect.ts
sBouzols: We cannot use Set here because we need to keep order to use lastSelectedElementId ...
My bad

src/utils/use-multiselect.ts
sBouzols: We cannot use Set here because we need to keep order to use lastSelectedElementId ...
My bad

src/utils/use-multiselect.ts
sBouzols: will do later

src/utils/use-multiselect.ts
klesaulnier: changed, even if I think it was not necessary

src/utils/use-multiselect.ts
klesaulnier: I think it was easier to read with "checkbox"
changed anyway

src/utils/use-multiselect.ts
klesaulnier: what if user does not want its component to implement all of these behaviours ? 
I think we should keep things simple for now, if we need this hook to be more powerful later, we'll change it

src/utils/use-multiselect.ts
klesaulnier: done

src/utils/use-multiselect.ts
klesaulnier: I think it's wrong, you're assuming we should always add an element to the selection if the element is right clicked
I don't think this is a common behaviour and it should not belong here

src/utils/use-multiselect.ts
klesaulnier: I prefer the way this component is made right now rather than an obscure method which handles everything for all possible clicks

sBouzols (approved):
Code Review OK
Test OK
Console warning check OK

EstherDarkish (approved):

PR #391, opened 2024-03-25
Fixed cursor positioning + invisible adornment

ne0ds:


src/components/dialogs/field-hook.js
```diff
@@ -141,19 +143,18 @@ export const useNameField = ({
 
     const debouncedUpdateValidity = useDebounce(updateValidity, 700);
 
-    useEffect(() => {
+    const adornment = useMemo(() => {
```
klesaulnier, line 146: nice, it's really better this way

klesaulnier (approved):

PR #402, opened 2024-04-05
Separate download and convert format for cases

SlimaneAmar:


src/components/utils/caseUtils.ts
```diff
@@ -194,5 +216,20 @@ export function useDownloadUtils() {
...
+            snackInfo({
+                messageTxt: buildPartialDownloadMessage(
+                    casesUuids.length,
+                    selectedElements
+                ),
+            });
+        }
+    };
+
+    return [handleDownloadCases, handleExportCases];
```
klesaulnier: return an object or make 2 differents hooks
here it would make it difficult for a dev to use handleExportCases without using handleDownloadCases

src/components/utils/caseUtils.ts
```diff
@@ -168,7 +190,7 @@ export function useDownloadUtils() {
             messageTxt: errorMsg,
         });
 
-    const handleDownloadCases = async (
+    const handleExportCases = async (
```
klesaulnier:
```suggestion
    const handleConvertCases = async (
```

to match the wording used in the UI
Otherwise it's difficult to understand what is different between downloading and exporting

src/components/toolbars/content-toolbar.js
```diff
@@ -216,10 +228,11 @@ const ContentToolbar = (props) => {
...
     }, [
         allowsDelete,
-        allowsDownloadCases,
+        allowsDownloadExportCases,
         allowsMove,
-        selectedElements.length,
-        stashedElements.length,
+        handleDownloadCases,
+        selectedElements,
+        stashedElements,
```
klesaulnier: this change could cause troubles, is it intended ?

src/components/utils/caseUtils.ts
klesaulnier: I think I prefer making a new hook, except intl / snackmessage, it does not have anything it common with existing hook

SlimaneAmar: Replaced by an  object

src/components/utils/caseUtils.ts
SlimaneAmar: Done

src/components/toolbars/content-toolbar.js
SlimaneAmar: Restored` stashedElements.lenght`

src/components/utils/caseUtils.ts
```diff
@@ -194,5 +216,20 @@ export function useDownloadUtils() {
...
+            snackInfo({
+                messageTxt: buildPartialDownloadMessage(
+                    casesUuids.length,
+                    selectedElements
+                ),
+            });
+        }
+    };
+
+    return { handleDownloadCases, handleConvertCases: handleConvertCases };
```
klesaulnier:
```suggestion
    return { handleDownloadCases, handleConvertCases };
```

klesaulnier (approved):

PR #412, opened 2024-04-23
Change delete confirmation message

antoinebhs:
Change delete message to:
* always have "Confirmation" as title
* not display the list of items we delete (we only display the selected to-delete directory as the focus is not very clear now)
* add "permanently" to emphasize the fact that the deletion is definitive


src/components/dialogs/delete-dialog.js
```diff
@@ -74,23 +74,8 @@ const DeleteDialog = ({
...
-                              tooltipSx={styles.tooltip}
-                          />
-                      ),
-                  }
-              )
-            : intl.formatMessage(
-                  { id: 'deleteMultipleItemsDialogTitle' },
-                  { itemsCount: items.length }
-              );
+    const buildTitle = () => {
```
TheMaskedTurtle, line 77: You can delete the argument in the method call

src/translations/fr.json
```diff
@@ -65,12 +65,11 @@
   "renameFolder": "Renommer",
   "insertNewDirectoryDialogTitle": "Créer un dossier",
   "insertNewRootDirectoryDialogTitle": "Créer un dossier racine",
-  "deleteDirectoryDialogMessage": "Le dossier {itemName} ainsi que les éléments qu'il contient va être supprimé.",
-  "deleteMultipleDirectoriesDialogMessage": "{itemsCount} dossiers ainsi que les éléments qu'ils contiennent vont être supprimés.",
-  "deleteItemDialogMessage": "{itemName} va être supprimé.",
-  "deleteMultipleItemsDialogMessage": "{itemsCount} éléments vont être supprimés.",
-  "deleteItemDialogTitle": "Supprimer\u00a0{itemName}",
-  "deleteMultipleItemsDialogTitle": "Supprimer {itemsCount} éléments",
+  "deleteDirectoryDialogMessage": "Le dossier {itemName} ainsi que les éléments qu'il contient va être supprimé définitivement.",
```
TheMaskedTurtle: "vont être supprimés" ?

TheMaskedTurtle (changes requested):

src/components/dialogs/delete-dialog.js
antoinebhs: You're right... Fixed!

src/translations/fr.json
antoinebhs: Fixed!

TheMaskedTurtle (approved):

PR #413, opened 2024-04-24
sequencing cases conversion/download

Mathieu-Deharbe:


src/components/utils/caseUtils.ts
```diff
@@ -200,19 +208,38 @@ export function useDownloadUtils() {
...
+            ) {
+                message += buildPartialDownloadMessage(
                     cases.length,
                     selectedElements
-                ),
-            });
+                );
+            }
+            if (message.length > 0) {
+                snackInfo({
```
thangqp, line 239: A small bug arrives after migrate the new version common ui, i.e. the snackInfo without headerId do not show the snack bar. headerId is optional as declared in the type definition file.
This bug will be corrected later in common ui

thangqp (approved):
Code OK, Test OK

src/components/utils/caseUtils.ts
```diff
@@ -31,48 +32,48 @@ const downloadCases = async (uuids: string[]) => {
...
+            link.href = href;
+            link.setAttribute('download', filename);
+            document.body.appendChild(link);
+            link.click();
+            document.body.removeChild(link);
+        } catch (error: any) {
+            if (error.name === 'AbortError') {
+                throw error;
+            }
+            handleCaseExportError?.(caseElement, error);
```
jamal-khey: why do use use '?' here ? handleCaseExportError seem to be always defined

may be move handleCaseExportError definition on the top of the function ?

jamal-khey (approved):
Test OK
No warning on the console

src/components/utils/caseUtils.ts
Mathieu-Deharbe: Indeed. Done.

src/components/utils/caseUtils.ts
Mathieu-Deharbe: Indeed. But I merge anyway : this is a priority correction and it still work despite the bug in main. The snackbar is simply not displayed at the end of the download (or its cancellation).

PR #415, opened 2024-04-24
Fix search component

ghazwarhili:


antoinebhs (approved):
Code OK
Test: there is still an issue when the user inputs a random string (e.g. "dfsgfdsreg") that does not exist in the index and hits enters, the search text is replaced by "undefined". This will be handled in a separate PR.

PR #416, opened 2024-04-24
First page : rootDirectory creation

souissimai:


src/components/directory-content.js
```diff
@@ -9,8 +9,10 @@ import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { setActiveDirectory, setSelectionForCopy } from '../redux/actions';
 import { FormattedMessage, useIntl } from 'react-intl';
+import { DialogsId } from '../utils/UIconstants';
```
AAJELLAL: You don't need this import. **constants** contains all elements . You can use it by calling  **constant.DialogsId**

src/components/directory-content.js
```diff
@@ -821,7 +855,35 @@ const DirectoryContent = () => {
             })),
         [childrenMetadata, currentChildren, getElementTypeTranslation]
     );
+    const handleOpenDialog = useCallback(() => {
+        setOpenDialog(DialogsId.ADD_ROOT_DIRECTORY);
+    }, [setOpenDialog]);
 
+    const renderNoContent = () => {
+        return (
+            <>
```
AAJELLAL: No need for this.

src/components/directory-content.js
```diff
@@ -949,8 +1011,10 @@ const DirectoryContent = () => {
         }
 
         // If no selection or currentChildren = null (first time) render nothing
-        // TODO : Make a beautiful page here
         if (!currentChildren || !selectedDirectory) {
+            if (treeData.rootDirectories.length === 0 && treeData.initialized) {
```
AAJELLAL: Why do you need the **initialized**? 
**treeData.rootDirectories.length === 0** is not sufficient in order to show renderNoContent?

src/components/tree-views-container.js
```diff
@@ -287,6 +287,7 @@ const TreeViewsContainer = () => {
                     setTreeData({
                         rootDirectories: nrs,
                         mapData: mdr,
+                        initialized: true,
```
AAJELLAL: the same here.

src/translations/fr.json
```diff
@@ -17,6 +17,7 @@
   "private": "privé",
   "public": "public",
   "emptyDir": "Dossier vide",
+  "firstDir": "Créer  votre premier dossier",
```
AAJELLAL: Remove the extra space plz.

AAJELLAL (changes requested):

src/components/directory-content.js
souissimai: yes need it because it takes time to fetch rootDirectories (as discussed)

src/components/directory-content.js
```diff
@@ -821,7 +854,31 @@ const DirectoryContent = () => {
...
+                        variant="contained"
+                        sx={styles.noContentButton}
+                        onClick={handleOpenDialog}
+                    >
+                        <FormattedMessage id={'createFolder'} />
+                    </Button>
+                </Box>
+            </Box>
+        );
+    };
```
klesaulnier: make a new component instead, in another file
this file is already big enough :)

src/components/directory-content.js
```diff
@@ -821,7 +854,31 @@ const DirectoryContent = () => {
...
+    const renderNoContent = () => {
+        return (
+            <Box sx={styles.noContentContainer}>
+                <Box sx={styles.noContentCircle}>
+                    <CreateNewFolderOutlinedIcon sx={styles.noContentIcon} />
+                </Box>
+                <Box sx={styles.noContentText}>
+                    <h1>
+                        <FormattedMessage id={'firstDir'} />
+                    </h1>
```
klesaulnier: this message is hard to see, change its color ?

src/components/directory-content.js
```diff
@@ -98,6 +98,38 @@ const styles = {
...
+        flexDirection: 'column',
+        justifyContent: 'center',
+        alignItems: 'center',
+        marginTop: theme.spacing(10),
+    }),
+    noContentCircle: (theme) => ({
+        position: 'relative',
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
```
klesaulnier: position: relative here and position: absolute for child component are doing the job, we don't need those 3 attributes

src/components/directory-content.js
```diff
@@ -98,6 +98,38 @@ const styles = {
         maxWidth: '250px',
         maxHeight: '50px',
     },
+    noContentContainer: (theme) => ({
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        alignItems: 'center',
```
klesaulnier: I don't think we need "alignItems: 'center'"

src/components/directory-content.js
```diff
@@ -98,6 +98,38 @@ const styles = {
...
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+        width: 200,
+        height: 200,
+        backgroundColor: theme.row.primary,
+        borderRadius: theme.spacing(15),
+    }),
+    noContentIcon: {
+        fontSize: '100px',
```
klesaulnier: I think I would make a component which only contain this icon is a circle

Then I would create a constant "CIRCLE_SIZE"
Then here, width and height would be equal to "CIRCLE_SIZE"
"fontSize" would be equals to "CIRCLE_SIZE / 2"

I would make it easier to reuse and isolate complex CSS

src/components/directory-content.js
```diff
@@ -766,7 +799,7 @@ const DirectoryContent = () => {
                 });
         }
         setSelectedUuids(new Set());
-    }, [handleError, currentChildren, currentChildrenRef]);
+    }, [handleError, currentChildren, currentChildrenRef, treeData]);
```
klesaulnier: which is this change necessary ?

src/components/directory-content.js
```diff
@@ -821,7 +854,31 @@ const DirectoryContent = () => {
             })),
         [childrenMetadata, currentChildren, getElementTypeTranslation]
     );
+    const handleOpenDialog = useCallback(() => {
+        setOpenDialog(constants.DialogsId.ADD_ROOT_DIRECTORY);
+    }, [setOpenDialog]);
```
klesaulnier: since "setOpenDialog" is a "useState" setter, you don't need to put it in dependencies

src/translations/en.json
```diff
@@ -17,6 +17,7 @@
   "private": "private",
   "public": "public",
   "emptyDir": "Empty directory",
+  "firstDir": "Create your first directory",
```
klesaulnier:
```suggestion
  "createFirstDir": "Create your first directory",
```

src/components/directory-content.js
klesaulnier: I think is would be cleaner with a "isLoading" functionnality, but it might be a little complex since treeData is is redux
here it sounds like a hack

klesaulnier (changes requested):

src/components/directory-content.js
souissimai: at the beginning I was thinking about adding isLoading functionality, but didnt find another way to do it since treedata is redux

src/components/directory-content.js
souissimai: done

src/components/directory-content.js
souissimai: color: theme.row.primary what do you suggest, comparing it with the provided image of stephane it's the same color desired.

src/components/directory-content.js
souissimai: unuseful to clean

src/components/directory-content.js
souissimai: Ithink we need it to make it on the center : image

src/translations/en.json
souissimai: done

src/components/directory-content.js
souissimai: done, refactored

src/components/directory-content.js
souissimai: Done

AAJELLAL (approved):
Tests and code are OK.
No warnings in console.

src/components/no-content-directory.tsx
```diff
@@ -0,0 +1,76 @@
...
+        noContentButton: {
+            borderRadius: '30px',
+        },
+    };
+
+    return (
+        <Box sx={styles.noContentContainer}>
+            <CircleIcon size={CIRCLE_SIZE} iconStyles={stylesIcon.circle}>
+                <CreateNewFolderOutlinedIcon
+                    style={{ fontSize: CIRCLE_SIZE / 2 }}
```
klesaulnier: use sx instead since it's MUI Icon

src/components/no-content-directory.tsx
```diff
@@ -0,0 +1,76 @@
...
+            alignItems: 'center',
+            marginTop: theme.spacing(20),
+        }),
+        noContentIcon: {
+            fontSize: `${CIRCLE_SIZE / 2}px`,
+            position: 'absolute',
+            top: '50%',
+            left: '50%',
+            transform: 'translate(-50%, -50%)',
+        },
```
klesaulnier: unused class, to remove

klesaulnier (approved):
still confused about the colors not being very easy to read, but nothing that can't be change easily

Test: OK
Code: OK

src/components/no-content-directory.tsx
souissimai: done

src/components/no-content-directory.tsx
souissimai: done

PR #421, opened 2024-04-30
Feed language to form context instead of Custom mui dialog

Meklo:
Also adds missing updateElement API call necessary for description modal to work


src/components/dialogs/contingency-list/creation/contingency-list-creation-dialog.jsx
```diff
@@ -53,10 +53,13 @@ const ContingencyListCreationDialog = ({ onClose, open, titleId }) => {
...
-    const methods = useForm({
-        defaultValues: emptyFormData,
-        resolver: yupResolver(schema),
-    });
+    const methods = {
+        ...useForm({
+            defaultValues: emptyFormData,
+            resolver: yupResolver(schema),
+        }),
+        language: languageLocal,
```
flomillot: Language is not a form method and shouldn't be mixed with.
You need to change a bit commons-ui instead, and not this part.

```
        <CustomFormProvider
            {...formMethods}
            validationSchema={formSchema}
            removeOptional={removeOptional}
            **language={language}**
        >
```

flomillot (changes requested):

src/components/dialogs/contingency-list/creation/contingency-list-creation-dialog.jsx
Meklo: Alright it is done, here is the corresponding PR [#440](https://github.com/gridsuite/commons-ui/pull/440)

flomillot (approved):

Meklo:
Waiting for next commons ui release before merge

PR #426, opened 2024-05-07
Handle access rights.

AAJELLAL:


src/components/dialogs/delete-dialog.jsx
```diff
@@ -50,12 +50,12 @@ const DeleteDialog = ({
     const openRef = useRef(null);
 
     useEffect(() => {
-        if (open && !openRef.current) {
+        if ((open && !openRef.current) || error !== '') {
```
SlimaneAmar, line 53: Why we show the dialog when an error occur (HttpResponseError : 403 Forbidden)
A snackbar is displayed

SlimaneAmar:
Remove `ACCESS_RIGHTS` in UIconstants.js

src/components/dialogs/delete-dialog.jsx
AAJELLAL: Unless I am mistaken, it was necessary to display the error message in both the pop-up and the snackbar since the snackbar disappears after a few seconds.
If it's not the case, I think the  component needs to be cleaned up. The error prop will no longer be useful. Check for example **ContentContextualMenu** :  a state is created and updated to manage the error message and pass it to the **delete-dialog**, but the message is never displayed.
The same thing goes for **DirectoryTreeContextualMenu**

SlimaneAmar (approved):

PR #427, opened 2024-05-13
Migrate directory content to AGgrid

Meklo:
Inspired by [this](https://github.com/gridsuite/gridexplore-app/pull/403) PR
Major changes :
- Usage of AG Grid instead of MUI Table for the directory content table : [here's](https://github.com/gridsuite/commons-ui/pull/434) the PR adding the needed component to commons-ui 
- CustomAGGrid component have been tweaked in order to use an alternate theme in order to better match material design
- The logic for retrieving and enriching the table's data from the has been moved to a dedicated hook : [useDirectoryContent.ts](https://github.com/gridsuite/gridexplore-app/pull/427/files#diff-a06aec2a79a13213a15207f30e50f53b34130db510fceffd113d755aee446919)
- Columns definitions have been moved to a dedicated utils file : [directory-content-utils.ts](https://github.com/gridsuite/gridexplore-app/pull/427/files#diff-cd4a55ae9c4a7063f66cebb7952659c5e68814973697fab66010aed354c6b42b)
- Renderers for each columns have been moved to a dedicated file : [directory-content-renderers.tsx](https://github.com/gridsuite/gridexplore-app/pull/427/files#diff-5d48504cc3712fd19b6310163e25526f79474f9c41a8831ac929de0fc8497945)
- The logic around the retrieval of rows selection - mainly needed for contextual menu and content toolbar - have been revamped in order to take advantage of AG Grid API


src/hooks/useDirectoryContent.ts
```diff
@@ -0,0 +1,84 @@
...
+import React, { useRef, useEffect, useCallback, useState } from 'react';
+import { useSnackMessage } from '@gridsuite/commons-ui';
+import { fetchElementsInfos } from '../utils/rest-api';
+import { UUID } from 'crypto';
+import { IElement, IElementMetadata, ReduxState } from '../redux/reducer.type';
+
+type useDirectoryContentReturnType = [
+    data: IElement[],
+    childrenMetadata: Record<UUID, IElementMetadata>
+];
```
klesaulnier: I think it's better to type "data" and "childrenMetadata" states instead
then return types would be automatically infered

src/hooks/useDirectoryContent.ts
```diff
@@ -0,0 +1,84 @@
...
+        let childrenToFetchElementsInfos = Object.values(currentChildren)
+            .filter((e) => !e.uploading)
+            .map((e) => e.elementUuid);
+        if (childrenToFetchElementsInfos.length > 0) {
+            fetchElementsInfos(childrenToFetchElementsInfos)
+                .then((res) => {
+                    res.forEach((e: IElementMetadata) => {
+                        metadata[e.elementUuid] = e;
+                    });
+                })
```
klesaulnier:
```suggestion
                .then((res: IElementMetadata[]) => {
                    res.forEach((e) => {
                        metadata[e.elementUuid] = e;
                    });
                })
```

src/utils/directory-content-renderers.tsx
klesaulnier, line 1: make a different file for each cell renderer to reduce file size and improve readability
maybe put them in a specific directory like "aggrid-cell-renderers"

src/components/utils/directory-content-utils.ts
```diff
@@ -0,0 +1,120 @@
...
+    UserCellRenderer,
+} from '../../utils/directory-content-renderers';
+import { IntlShape } from 'react-intl';
+import { UUID } from 'crypto';
+import { IElementMetadata } from '../../redux/reducer.type';
+import { AgGridReact } from 'ag-grid-react';
+import React from 'react';
+
+export const formatMetadata = (
+    data: any,
```
klesaulnier:
```suggestion
    data: IElementMetadata,
```

src/components/utils/directory-content-utils.ts
```diff
@@ -0,0 +1,120 @@
...
+});
+
+export const computeCheckedElements = (
+    gridRef: React.MutableRefObject<AgGridReact | null>,
+    childrenMetadata: Record<UUID, IElementMetadata>
+) => {
+    return (
+        gridRef.current?.api
+            ?.getSelectedRows()
+            .map((row) => formatMetadata(row, childrenMetadata)) ?? []
```
klesaulnier:
```suggestion
            .map((row: IElementMetadata) => formatMetadata(row, childrenMetadata)) ?? []
```

src/components/app-wrapper.jsx
```diff
@@ -94,6 +95,8 @@ let lightTheme = createTheme({
             textTransform: 'none',
         },
     },
+    aggridMaterialColor: '#90caf9 !important',
+    aggridHiglightColor: '#8e9c9b !important',
```
klesaulnier: where does those color codes come from ?

src/components/app-wrapper.jsx
```diff
@@ -148,6 +152,8 @@ let darkTheme = createTheme({
             textTransform: 'none',
         },
     },
+    aggridMaterialColor: '#1976d2 !important',
+    aggridHiglightColor: '#545c5b !important',
```
klesaulnier: same question here

src/components/directory-content.jsx
```diff
@@ -62,28 +63,12 @@ const styles = {
...
-    cell: {
-        display: 'flex',
-        alignItems: 'center',
-        textAlign: 'center',
-        boxSizing: 'border-box',
-        flex: 1,
-        height: '48px',
-        padding: `${DEFAULT_CELL_PADDING}px`,
-        overflow: 'hidden',
+    grid: { height: 500, width: '100%' },
```
klesaulnier: height: 500, why do we need this ?

src/components/directory-content.jsx
```diff
@@ -62,28 +63,12 @@ const styles = {
...
+    overflow: {
+        whiteSpace: 'pre',
         textOverflow: 'ellipsis',
+        overflow: 'hidden',
     },
```
klesaulnier: unused property

src/components/directory-content.jsx
```diff
@@ -62,28 +63,12 @@ const styles = {
...
-    },
-    icon: (theme) => ({
-        marginRight: theme.spacing(1),
-        width: '18px',
-        height: '18px',
-    }),
-    circularRoot: (theme) => ({
-        marginRight: theme.spacing(1),
-    }),
     checkboxes: {
```
klesaulnier: unused property

src/hooks/useDirectoryContent.ts
```diff
@@ -0,0 +1,84 @@
...
+        React.SetStateAction<boolean>
+    >
+): useDirectoryContentReturnType => {
+    const currentChildren = useSelector(
+        (state: ReduxState) => state.currentChildren
+    );
+    const [childrenMetadata, setChildrenMetadata] = useState({});
+    const { snackError } = useSnackMessage();
+
+    const [data, setData] = useState(currentChildren);
```
klesaulnier: rename data to something more specific ? here it's hard to say what we are returning

src/components/directory-content.jsx
```diff
@@ -293,93 +263,52 @@ const DirectoryContent = () => {
...
 
-    /* User interactions */
-    const contextualMixPolicies = useMemo(
-        () => ({
-            BIG: 'GoogleMicrosoft', // if !selectedUuids.has(selected.Uuid) deselects selectedUuids
-            ALL: 'All', // union of activeElement.Uuid and selectedUuids (currently implemented)
-        }),
-        []
-    );
-    const contextualMixPolicy = contextualMixPolicies.ALL;
```
klesaulnier: why did you remove this ?
also, I think the way it worked changed

before, when selecting several elements, when we right clicked another element, this element was added to the selection, it's not visually the case anymore (even if it seems to work)

src/components/directory-content.jsx
```diff
@@ -293,93 +263,52 @@ const DirectoryContent = () => {
...
-            currentChildren,
-            dispatch,
-            selectedDirectory,
-            selectedUuids,
-            contextualMixPolicies,
-            contextualMixPolicy,
-            childrenMetadata,
-        ]
+        [dispatch, selectedDirectory]
     );
```
klesaulnier: is it to handle right click on headers ? 
Or is it to prevent multiple contextual menus to open when right clicking on aggrid ?
It sounds complicated with the "isRowHovered", and it does not seem to work properly (check last comment about bugs)

src/components/directory-content.jsx
```diff
@@ -412,434 +341,142 @@ const DirectoryContent = () => {
...
         },
         [
             childrenMetadata,
-            currentChildren,
             dispatch,
             getLink,
             handleError,
             intl,
-            selectedDirectory?.elementUuid,
+            selectedDirectory,
```
klesaulnier:
```suggestion
            selectedDirectory?.elementUuid,
```
is it a necessary change ? we only need selectedDirectory?.elementUuid here

src/components/directory-content.jsx
```diff
@@ -412,434 +341,142 @@ const DirectoryContent = () => {
...
         setIsMissingDataAfterDirChange(true);
     }, [selectedDirectory, setIsMissingDataAfterDirChange]);
 
-    useEffect(() => {
-        if (!currentChildren?.length) {
-            setChildrenMetadata({});
-            setIsMissingDataAfterDirChange(false);
-            return;
-        }
+    const [checkedElement, setCheckedElement] = useState([]);
```
klesaulnier: create states at the beginning of components when possible

src/components/directory-content.jsx
```diff
@@ -412,434 +341,142 @@ const DirectoryContent = () => {
...
-            })),
-        [childrenMetadata, currentChildren, getElementTypeTranslation]
-    );
+        return selection;
+    }, [
+        activeElement,
+        checkedElement,
+        childrenMetadata,
+        isActiveElementUnchecked,
+    ]);
```
klesaulnier: currently, right clicking an element will add it to the selection
I think it will simplify the code if we keep this behaviour

src/components/directory-content.jsx
```diff
@@ -875,87 +505,42 @@ const DirectoryContent = () => {
...
+        if (
+            ![ElementType.CASE, ElementType.PARAMETERS].includes(
+                cellData.data?.type
+            )
+        ) {
+            style.cursor = 'pointer';
+        }
+        return style;
+    }, []);
+
```
klesaulnier: can all those methods be created out of the component instead ? It would prevent the usage of useCallback

src/components/directory-content.jsx
```diff
@@ -993,6 +578,7 @@ const DirectoryContent = () => {
                         setActiveElement(null);
                         setOpenDescModificationDialog(false);
                     }}
+                    updateElement={updateElement}
```
klesaulnier: I don't understand, does that mean it did not work before ?

klesaulnier (changes requested):
Some bugs detected
- right clicking an element twice will create multiple contextual menus
- I have a case where a study is stuck in "creation in progress" : before, the type was empty for them, now I have an untranslated "FILTER" label displayed, which causes warning in console
- editing a description now prompts this warning in console : Warning: React does not recognize the `adornmentPosition` prop on a DOM element
- when using the app in french, a enormous amount of translation warning appear in console

Tests: OK 

Need to check : 
- alternative theme
- editing description button is now working on the whole column and can confuse user

flomillot:
Don't forget to delete the react-virtualized dep added by this PR : https://github.com/gridsuite/gridexplore-app/pull/429
Also check that it's removed from package-lock.

klesaulnier:
There is a bug where there are many elements in a directory
A scroll bar appears, but the bottom elements can't be seen

src/hooks/useDirectoryContent.ts
Meklo: Done

src/hooks/useDirectoryContent.ts
Meklo: Done

src/components/app-wrapper.jsx
Meklo: Those were extracted from the current implementation

src/components/app-wrapper.jsx
Meklo: See above

src/components/directory-content.jsx
Meklo: Removed

src/components/directory-content.jsx
Meklo: Removed

src/hooks/useDirectoryContent.ts
Meklo: I renamed it to `rows`

src/components/directory-content.jsx
Meklo: I thought it was partly the cause of the abusive rerender issue I was having but reinstating it doesn't seem to break things down so fine by me

src/components/directory-content.jsx
Meklo: Done, renamed it checkedRows by the way

src/components/directory-content.jsx
Meklo: Yup, wasn't supposed to be in the PR and it was fixed it the meantime

src/components/directory-content.jsx
Meklo: I was initially having issues with the table collapsing, doesn't seem to be needed anymore

src/utils/directory-content-renderers.tsx
Meklo: Done

src/components/utils/directory-content-utils.ts
Meklo: Done

src/components/utils/directory-content-utils.ts
Meklo: Done

src/components/directory-content.jsx
Meklo: I wasn't sure it was still needed, I put it back and adapted it to work with AGgrid

src/components/directory-content.jsx
Meklo: It handles context menu event on the container wrapping the grid
To prevent the bug I now handle every context menu event through this one instead of using AG grid built in cell context menu event. 
There's no visual glitches anymore and the only drawback is when you try to open a row context menu while another one is already open it will open the general context menu

src/components/directory-content.jsx
Meklo: The feature is now back, not sure how it would simplify the code though since it needs additional logic

src/components/directory-content.jsx
Meklo: I externalized those in a CustomAGGgrid wrapper dedicated for the directory content, it's called DirectoryContentTable.
It reduces boilerplate code inside directory content imo, wdyt ?

Meklo:
> There is a bug where there are many elements in a directory A scroll bar appears, but the bottom elements can't be seen

There was a weird issue where AG grid would consider its total size including the content toolbar component above the grid so the toolbar height was deduced from the grid's height, reworking the DOM hierarchy fixed this problem

Meklo:
Regarding `adornmentPosition` warning it was indeed recently introduced by commons ui changes but it is not related to the changes at hand
Regarding the warning related to french translation it is the case for all AG grid usages across gridexplore and gridstudy and is deliberate if I'm not mistaken. When I did the first AG grid implementation those translations were filled to prevent the said warnings but since they are about elements of the UI we don't use it has been decided to remove those translation. Maybe it's not the whole picture but It's where I'm at on this subject 

The rest is normally fixed

src/components/app-wrapper.jsx
klesaulnier: do you mean we have the same code in gridstudy ?

src/components/DirectoryContentTable.tsx
```diff
@@ -0,0 +1,70 @@
...
+    handleRowSelected: () => void;
+    handleCellClick: () => void;
+    colDef: ColDef[];
+}
+
+const onGridReady = ({ api }: GridReadyEvent<IElement>) => {
+    api?.sizeColumnsToFit();
+};
+
+const getRowId = (params: GetRowIdParams<IElement>): string =>
```
klesaulnier:
```suggestion
const getRowId = (params: GetRowIdParams<IElement>) =>
```
No need to type return type when it can be inferred
Here you are even changing "UUID" to "string" type by doing so

achour94 (commented):
Some Remarks:
We lost the pointer on the row checkbox selection
When we right click on a row multiple times, it does not shows the contextual menu related to the row from the second click (I don't know if we can fix that ! )

src/components/DirectoryContentTable.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+import { defaultColumnDefinition } from './utils/directory-content-utils';
+import { CustomAGGrid, ElementType } from '@gridsuite/commons-ui';
+import { AgGridReact } from 'ag-grid-react';
+import { IElement } from '../redux/reducer.type';
+import { GetRowIdParams } from 'ag-grid-community/dist/types/core/interfaces/iCallbackParams';
+import { ColDef, GridReadyEvent, RowClassParams } from 'ag-grid-community';
+
+interface DirectoryContentTableProps {
+    gridRef: React.ForwardedRef<AgGridReact<IElement>>;
```
klesaulnier:
```suggestion
    gridRef: RefObject<AgGridReact<IElement>>;
```

src/components/directory-content.jsx
```diff
@@ -304,83 +266,84 @@ const DirectoryContent = () => {
...
+                            ]);
+                            gridRef.current?.api.setNodesSelected({
+                                nodes: [
+                                    ...gridRef.current?.api.getSelectedNodes(),
+                                    gridRef.current?.api.getRowNode(
+                                        row.data.elementUuid
+                                    ),
+                                ],
+                                newValue: true,
+                            });
```
klesaulnier:
```suggestion
                            gridRef.current?.api.getRowNode(
                                        row.data.elementUuid
                            ).setSelected(true),
```

src/components/directory-content.jsx
```diff
@@ -304,83 +266,84 @@ const DirectoryContent = () => {
...
-                            !selectedUuids.has(element.elementUuid)
-                        ) {
-                            let updatedSelectedUuids = new Set(selectedUuids);
-                            updatedSelectedUuids.add(element.elementUuid);
-                            setSelectedUuids(updatedSelectedUuids);
+                        if (isRowUnchecked(row, checkedRows)) {
+                            setCheckedRows((prevCheckedRows) => [
+                                ...prevCheckedRows,
+                                formatMetadata(row.data, childrenMetadata),
+                            ]);
```
klesaulnier: is it necessary ? I think "handleRowSelected" is called when row selection is changing and should be done already

src/components/directory-content.jsx
```diff
@@ -304,83 +266,84 @@ const DirectoryContent = () => {
...
                     if (contextualMixPolicy === contextualMixPolicies.BIG) {
                         // If some elements were already selected and the active element is not in them, we deselect the already selected elements.
-                        if (
-                            selectedUuids?.size &&
-                            element?.elementUuid &&
-                            !selectedUuids.has(element.elementUuid)
-                        ) {
-                            setSelectedUuids(new Set());
+                        if (isRowUnchecked(row, checkedRows)) {
+                            setCheckedRows([]);
```
klesaulnier: same here, it should be done already ?

src/components/directory-content.jsx
```diff
@@ -413,72 +376,96 @@ const DirectoryContent = () => {
...
+            if (event.colDef.field === 'description') {
+                handleDescriptionIconClick(event);
+            } else {
+                if (childrenMetadata[event.data.elementUuid] !== undefined) {
+                    setElementName(
+                        childrenMetadata[event.data.elementUuid]?.elementName
+                    );
+                    const subtype =
+                        childrenMetadata[event.data.elementUuid]
+                            .specificMetadata.type;
```
klesaulnier, line 372: change from "subtype" to "type"
does that mean it was not working before ?

src/components/directory-content.jsx
```diff
@@ -487,360 +474,44 @@ const DirectoryContent = () => {
...
-                                childrenMetadata[child.elementUuid]?.subtype,
-                            hasMetadata:
-                                childrenMetadata[child.elementUuid] !==
-                                undefined,
-                            ...child,
-                        };
-                    });
-            }
+    const handleRowSelected = useCallback(() => {
+        setCheckedRows(computeCheckedElements(gridRef, childrenMetadata));
```
klesaulnier, line 475: why do we need to add children metadata here ?
it is because we need them to display tooltips or not ?

src/components/directory-content.jsx
```diff
@@ -487,360 +474,44 @@ const DirectoryContent = () => {
...
         }
-        return [...new Set(selectedChildren)];
-    };
+        return selection;
+    }, [
+        activeElement,
+        checkedRows,
+        childrenMetadata,
+        isActiveElementUnchecked,
+    ]);
```
klesaulnier, line 490: I don't think all of this is necessary since you now add active element to selection when right clicking
after a few tests I think you can remove it and isActiveElementUnchecked

This is what I meant when I said it would simplify the code to keep old behaviour with contextMenuPolicy

src/components/directory-content.jsx
```diff
@@ -1110,17 +696,10 @@ const DirectoryContent = () => {
...
         <>
-            <div
-                style={{
-                    display: 'flex',
-                    flexDirection: 'column',
-                    flexGrow: 1,
-                }}
-                onContextMenu={(e) => onContextMenu(e)}
-            >
+            {rows && <ContentToolbar selectedElements={checkedRows} />}
```
klesaulnier, line 671: include this toolbar in below grid so we can use right click here just like in main branch ?

src/hooks/useDirectoryContent.ts
```diff
@@ -0,0 +1,82 @@
...
+) => {
+    const currentChildren = useSelector(
+        (state: ReduxState) => state.currentChildren
+    );
+    const [childrenMetadata, setChildrenMetadata] = useState<
+        Record<UUID, IElementMetadata>
+    >({});
+    const { snackError } = useSnackMessage();
+
+    const [rows, setRows] = useState<IElement[]>(currentChildren);
```
klesaulnier: what is the point of this state since it seems to be equals to currentChildren at all time ?

klesaulnier (commented):
rename all new files to match other files name format : xxxx-xxxxx.tsx
I'm confused with the right clicking behaviour, if a user right click on the wrong element and tries to right click on another one, it could lead to annoying behaviour -> we should check this with PO. But I agree it seems to be something quite hard to deal with technically


I agree with the warning, they seem to be here already everywhere else
All other bugs seem to be fixed after some tests

src/hooks/useDirectoryContent.ts
Meklo: It was initially to follow documentation guidelines but looking back at it it seems unnecessary bloating, I'll directly return currentChildren I think

src/components/directory-content.jsx
Meklo: Moving it into the Grid breaks everything down and I wasn't able to find another workaround as this one

src/components/directory-content.jsx
Meklo: Oh I see, I'll proceed to get rid of this

src/components/directory-content.jsx
Meklo: They are needed for the content toolbar as well as the contextual menu, they could be directly passed down those components or built in the datasource of the table to avoid complexity and confusion but that would require a substantial refactoring I reckon

src/components/directory-content.jsx
Meklo: How so ? 
I don't see the problem here, types are FILTER, STUDY, CONTINGENCY LIST, etc... and subtypes are mainly related to filters and contingency lists for CRITERIA_BASED, EXPLICIT_NAMING, EXPERT, etc...

src/components/directory-content.jsx
Meklo: Oof big improvement here, my solution was needlessly convoluted

src/components/DirectoryContentTable.tsx
Meklo: Done

src/components/DirectoryContentTable.tsx
Meklo: Done

src/components/directory-content.jsx
Meklo: Not needed indeed, simplifies a lot

src/components/directory-content.jsx
Meklo: Done

src/components/directory-content.jsx
Meklo: It was too good to be true, the issue here is the use case when we open the contextual menu without having previously checked any rows, the contextual policy does not add the active element to the selection

Meklo: I guess it could be fixed and simplified by a deep refactoring of both the content toolbar and the contextual menu but I'd avoid bulldozing everything for this PR

src/components/directory-content.jsx
klesaulnier: since it changes the behaviour, we should ask the PO if this change is OK
we could add a box wrapping both which would hold the "onContextMenu" handler

klesaulnier (approved):

achour94:
The custom aggrid table in RHF section does not render correctly

achour94 (approved):

PR #441, opened 2024-06-07
Add csv config to contingency list creation

YenguiSeddik:


jamal-khey (approved):
Code Ok
Console logs Ok
Test Ok

src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.jsx
```diff
@@ -182,6 +182,7 @@ const ExplicitNamingForm = () => {
                 fileHeaders: csvFileHeaders,
                 getDataFromCsv: getDataFromCsvFile,
                 csvData: csvInitialData,
+                config: { delimiter: ',' },
```
sBouzols: Why fixing the delimiter to ',' ?
We should used `delimitersToGuess` prop and juste remove the one we use for multiple values here '|'
see https://react-papaparse.js.org/docs#config : 
```diff
-delimitersToGuess: [',', '	', '|', ';', RECORD_SEP, UNIT_SEP]
+delimitersToGuess: [',', '	', ';', RECORD_SEP, UNIT_SEP]
```

sBouzols: And we could fix this rule directly in commons-ui to avoid knowing papaparse here.

sBouzols (changes requested):
do not merge this PR.
Must be fixed for all CSV parse workflow then fix it in commons-ui

jamal-khey:
delimiter config has been moved to commun-ui and merged.
this pr is not usefull any more

PR #445, opened 2024-06-12
Fix SearchBar: unconsistent Tree when selecting element

sBouzols:


src/components/search/search-bar.tsx
klesaulnier, line 1: this is working nicely, I think we can improve one last thing
the directory we are redirected to is "fetched" twice, once because of "fetchDirectoryContent", and once because of "handleDispatchDirectory"
maybe we could remove last element from the "for (const uuid of elementUuidPath) {" loop to prevent this ?

I tried to do it quickly without success though

klesaulnier (approved):

src/components/search/search-bar.tsx
```diff
@@ -148,37 +148,36 @@ export const SearchBar: FunctionComponent<SearchBarProps> = ({ inputRef }) => {
...
-                                ) as IDirectory[]
-                            );
-                        })
-                        .catch((error) =>
-                            snackError({
-                                messageTxt: error.message,
-                                headerId: 'pathRetrievingError',
-                            })
+                try {
+                    for (const uuid of elementUuidPath) {
```
antoinebhs, line 161: The loop should be on all elements of the path except the last otherwise we have a double request on the element.

antoinebhs (approved):
Above comment will be fixed in a separate ticket.

src/components/search/search-bar.tsx
sBouzols: If we don't fetch the last one, then `handleDispatchDirectory` and following code will check first if selectedDirectory is in the map. If not, it will avoid select current directory.
By the way, selecting the current directory proc another fetchDirectoryContent then we have 0 or two identical fetch.
It was already like that, will refactor later.

PR #449, opened 2024-06-14
handle user quotas errors

AbdelHedhili:


src/components/dialogs/create-case-dialog/create-case-dialog.tsx
```diff
@@ -92,7 +95,16 @@ const CreateCaseDialog: React.FunctionComponent<CreateCaseDialogProps> = ({
...
+                if (
+                    err.status === 403 &&
+                    err.message.includes(HTTP_MAX_ELEMENTS_EXCEEDED_MESSAGE)
+                ) {
+                    let limit = err.message.split(/[: ]+/).pop();
+                    snackError({
+                        messageId: 'maxElementExceededError',
+                        messageValues: { limit: limit },
+                    });
+                } else if (err?.status === HTTP_UNPROCESSABLE_ENTITY_STATUS) {
```
achour94, line 98: extract this logic into a util function
```
import { HTTP_MAX_ELEMENTS_EXCEEDED_MESSAGE } from 'utils/UIconstants';

interface CustomError extends Error {
    status?: number;
}

export const handleMaxElementsExceededError = (
    error: CustomError,
    snackError: Function
): boolean => {
    if (
        error.status === 403 &&
        error.message.includes(HTTP_MAX_ELEMENTS_EXCEEDED_MESSAGE)
    ) {
        let limit = error.message.split(/[: ]+/).pop();
        snackError({
            messageId: 'maxElementExceededError',
            messageValues: { limit: limit },
        });
        return true;
    }
    return false;
};
```
and use it everywhere
```
if (handleMaxElementsExceededError(err, snackError)) {
                    return;
                }
```

achour94 (approved):

FranckLecuyer (approved):

PR #452, opened 2024-06-18
Rename the downloaded file when exporting or converting case

souissimai:


src/components/utils/caseUtils.ts
```diff
@@ -11,21 +11,30 @@ import {
...
+} from '@gridsuite/commons-ui';
 import { useCallback, useState } from 'react';
 
-const downloadCases = async (uuids: string[]) => {
-    for (const uuid of uuids) {
-        const result = await downloadCase(uuid);
-        let name = await getCaseOriginalName(uuid);
+const downloadCases = async (selectedCases: ElementAttributes[]) => {
+    for (const selectedCase of selectedCases) {
+        const result = await downloadCase(selectedCase.elementUuid);
```
SlimaneAmar, line 23: the fetch `downloadCase` need to get also the format so as not to call the `getCaseOriginalName`
In an another US

src/components/utils/caseUtils.ts
```diff
@@ -65,12 +74,16 @@ export function useDownloadUtils() {
                 .get('Content-Disposition')
                 .split('filename=')[1];
             filename = filename.substring(1, filename.length - 1); // We remove quotes
+            const fileExtension = filename.split('.').pop();
```
SlimaneAmar: Add an enpoint like` network-converion-server` with filename (without extension)
and get extension from the back

souissimai: we do not have extension, we have only format which is not the same: 
Example 
 format : MATPOWER
 extension : mat
I think that the actual solution to get format is to extract it from original case name

souissimai: done

src/components/utils/caseUtils.ts
```diff
@@ -57,20 +66,21 @@ export function useDownloadUtils() {
...
+                caseElement.elementName,
                 format,
                 formatParameters,
                 abortController
             );
             let filename = result.headers
                 .get('Content-Disposition')
                 .split('filename=')[1];
-            filename = filename.substring(1, filename.length - 1); // We remove quotes
+
```
SlimaneAmar: Need to remove quotes

src/components/menus/content-contextual-menu.jsx
```diff
@@ -350,6 +351,14 @@ const ContentContextualMenu = (props) => {
                     nameItem: elementName[1],
                 });
             }
+            //update selected elements if element is renamed
+            const updatedSelectedElements = selectedElements.map((element) => {
```
SlimaneAmar, line 355: Remove callback

src/components/utils/caseUtils.ts
souissimai: done

src/components/menus/content-contextual-menu.jsx
souissimai: done

SlimaneAmar (approved):

PR #462, opened 2024-06-25
Directory content table does not reload when searching element within same directory

klesaulnier:


souissimai (approved):

jamal-khey (approved):
Revu Ok 
console warning Ok

PR #466, opened 2024-06-28
Add 'not exhaustive' option to searchbar

klesaulnier:


src/components/search/search-bar.tsx
```diff
@@ -145,31 +154,64 @@ export const SearchBar: FunctionComponent<SearchBarProps> = ({ inputRef }) => {
...
+                ...other
+            }: {
+                children: React.ReactNode;
+            }) => {
+                return (
+                    <Paper {...other}>
+                        <Typography
+                            variant="body1"
+                            style={{
+                                color: theme.palette.info.main,
```
AAJELLAL: Proposition: Move the style definition to the top of the class.

jamal-khey: i don't know it is possible to do in this case, because i am using the theme that comes from a custum hook

klesaulnier (commented):
Approved

src/components/search/search-bar.tsx
jamal-khey: fixed by Kevin

AAJELLAL (approved):
Tests and code are OK.
No console errors

jamal-khey (approved):

PR #468, opened 2024-07-04
Fix the app crash when deleting a selected root folder.

AAJELLAL:


sBouzols (approved):
Code Review OK
Tests OK
Console Warning Check OK

PR #474, opened 2024-07-17
Use commons-ui v0.63.1

etiennehomer:


thangqp (approved):
Test some basic functionalities, expert filter, create study, copy study => OK

PR #475, opened 2024-07-19
Migrate reducer to TypeScript

Tristan-WorkGH:


src/components/app-top-bar.tsx
```diff
@@ -41,7 +51,7 @@ const AppTopBar = ({ user, userManager }) => {
 
     const [showParameters, setShowParameters] = useState(false);
 
-    const searchInputRef = useRef(null);
+    const searchInputRef = useRef<any | null>(null);
```
Meklo, line 54: Wouldn't `any` include `null` ?

src/components/app-top-bar.tsx
```diff
@@ -93,7 +103,9 @@ const AppTopBar = ({ user, userManager }) => {
                 globalVersionPromise={() =>
                     fetchVersion().then((res) => res?.deployVersion)
                 }
-                additionalModulesPromise={getServersInfos}
+                additionalModulesPromise={
+                    getServersInfos as () => Promise<GridSuiteModule[]>
```
Meklo, line 107: Is this cast necessary ? It compiles when I remove it

src/components/app-top-bar.tsx
```diff
@@ -103,11 +115,9 @@ const AppTopBar = ({ user, userManager }) => {
             />
         </>
     );
-};
+}
 
 AppTopBar.propTypes = {
     user: PropTypes.object,
```
Meklo, line 121: You can remove this one

src/components/search/search-bar.tsx
```diff
@@ -22,15 +22,16 @@ import {
...
 import {
+    AppState,
     ElementAttributesES,
     IDirectory,
     ITreeData,
-    ReduxState,
-} from '../../redux/reducer.type';
-import { RenderElementProps } from '@gridsuite/commons-ui/dist/components/ElementSearchDialog/element-search-input';
+} from '../../redux/reducer';
+import { ElementSearchInputProps } from '@gridsuite/commons-ui/dist/components/ElementSearchDialog/element-search-input';
```
Meklo: Shouldn't it be exported from commons index ?

src/components/search/search-bar.tsx
```diff
@@ -94,6 +97,7 @@ export const SearchBar: FunctionComponent<SearchBarProps> = ({ inputRef }) => {
                 setTreeData({
                     rootDirectories: newRootDirectories,
                     mapData: newMapData,
+                    initialized: true,
```
Meklo, line 108: Why this addition ?

src/redux/reducer.ts
```diff
@@ -0,0 +1,350 @@
...
+    subdirectoriesCount: number;
+    lastModificationDate: string;
+    pathName: string[];
+    pathUuid: UUID[];
+}
+
+export interface ITreeData {
+    rootDirectories: IDirectory[];
+    mapData: Record<string, IDirectory>;
+    initialized: boolean;
```
Meklo, line 98: How come this addition ?

Meklo (changes requested):

src/components/app-top-bar.tsx
Tristan-WorkGH: yes, that was to be clear for the next dev that we known it can be `null` and an unknown type

src/components/app-top-bar.tsx
Tristan-WorkGH: yes because `getServersInfos()` comme from a js file and return automatically `any` for typescript

src/components/search/search-bar.tsx
Tristan-WorkGH: mandatory / not optional field

src/redux/reducer.ts
Tristan-WorkGH: the reducer was incomplete/partial, I was obligated to add missing types

src/components/search/search-bar.tsx
Meklo: from which component point of view, redux or treeView ?

Tristan-WorkGH: From redux point of view, because in `ITreeData` the field isn't marked as optional.

Dunno if it can be or not from treeview side.

Meklo (approved):

Meklo (changes requested):
Missing licences in `module-emotion.d.ts` and `module-mui.d.ts`

Meklo (approved):

achour94 (approved):

PR #480, opened 2024-07-30
Case Export with choice of format: editing the file name

Mathieu-Deharbe:
allows the user to pick a filename in the case conversion dialog (export in a picked file format)


src/components/dialogs/export-case-dialog.tsx
```diff
@@ -19,13 +19,15 @@ import {
...
+    TextField,
     Typography,
 } from '@mui/material';
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { getExportFormats } from '../../utils/rest-api';
 import { FormattedMessage, useIntl } from 'react-intl';
 import { ExpandLess, ExpandMore } from '@mui/icons-material';
 import { CancelButton, FlatParameters } from '@gridsuite/commons-ui';
+import { ElementAttributes } from '@gridsuite/commons-ui/dist/utils/types';
```
thangqp, line 30: Mistake while importing

src/components/dialogs/export-case-dialog.tsx
```diff
@@ -203,4 +231,8 @@ const ExportCaseDialog = (props: ExportCaseDialogProps) => {
     );
 };
 
+ExportCaseDialog.defaultProps = {
+    selectedElements: [],
+};
+
```
thangqp: Not need defaultProps in TSX file

src/components/dialogs/export-case-dialog.tsx
```diff
@@ -109,6 +123,20 @@ const ExportCaseDialog = (props: ExportCaseDialogProps) => {
...
             <DialogContent>
+                {oneFileMode && (
+                    <TextField
+                        key="fileName"
+                        margin="dense"
+                        label={<FormattedMessage id="download.fileName" />}
+                        variant="filled"
+                        id="fileName"
+                        value={fileName}
+                        style={{ width: '100%' }}
```
thangqp: Using sx instead of style
and lack of a space, eg. margin bottom 8px, to below component

src/components/dialogs/export-case-dialog.tsx
```diff
@@ -46,14 +48,26 @@ type FormatParameters = {
...
+        fileName: string
+    ) => Promise<void>;
 }
 
 const ExportCaseDialog = (props: ExportCaseDialogProps) => {
     const [loading, setLoading] = useState<boolean>(false);
     const [formats, setFormats] = useState<ExportFormats>([]);
     const [selectedFormat, setSelectedFormat] = useState<string | null>(null);
+    const oneFileMode: boolean = useMemo(
+        () => props.selectedElements.length === 1,
```
thangqp: as convention, we now do destructuring at the signature of component, no longer using props.xxx except develop a wrapper component

src/components/dialogs/export-case-dialog.tsx
thangqp: @gridsuite/commons-ui/dist/utils/types
Mistake while importing

src/components/dialogs/export-case-dialog.tsx
```diff
@@ -109,6 +123,20 @@ const ExportCaseDialog = (props: ExportCaseDialogProps) => {
...
+                {oneFileMode && (
+                    <TextField
+                        key="fileName"
+                        margin="dense"
+                        label={<FormattedMessage id="download.fileName" />}
+                        variant="filled"
+                        id="fileName"
+                        value={fileName}
+                        style={{ width: '100%' }}
+                        onChange={(event: any) =>
```
thangqp: to remove type any

src/components/utils/caseUtils.ts
```diff
@@ -61,26 +61,31 @@ export function useDownloadUtils() {
...
+        abortController: AbortController,
+        fileName: string
     ): Promise<void> => {
         try {
+            // if no fileName is provided, the case name will be used
+            const autoName: boolean = fileName === undefined || fileName === '';
             const result = await fetchConvertedCase(
                 caseElement.elementUuid,
-                caseElement.elementName,
+                autoName ? caseElement.elementName : fileName,
```
thangqp: for default value in JS we can use || as following
fileName || caseElement.elementName

src/components/utils/caseUtils.ts
```diff
@@ -61,26 +61,31 @@ export function useDownloadUtils() {
...
                 format,
                 formatParameters,
                 abortController
             );
-            let filename = result.headers
+
+            fileName = result.headers
```
thangqp: avoid to reset a value for an argument, if name conflict, we can use another variable, eg. downloadFileName

src/components/utils/caseUtils.ts
```diff
@@ -225,7 +231,13 @@ export function useDownloadUtils() {
             setAbortController(controller);
 
             for (const c of cases) {
-                await exportCase(c, format, formatParameters, controller);
+                await exportCase(
+                    c,
+                    format,
+                    formatParameters,
+                    controller,
+                    fileName
```
thangqp: In my opinion, passing the same fileName for every cases is not a good idea, in fact the fileName you pass is only fileName of the first element.

I created a suggestion PR here which uses a map between case uuid and file name for each case: https://github.com/gridsuite/gridexplore-app/pull/489

In this PR, I have corrected also comments mentioned in this PR..

src/components/dialogs/export-case-dialog.tsx
Mathieu-Deharbe: OK

src/components/dialogs/export-case-dialog.tsx
Mathieu-Deharbe: OK

src/components/dialogs/export-case-dialog.tsx
Mathieu-Deharbe: OK

src/components/dialogs/export-case-dialog.tsx
Mathieu-Deharbe: Indeed sorry.

src/components/dialogs/export-case-dialog.tsx
Mathieu-Deharbe: As you wish. But what difference does it make ?

src/components/utils/caseUtils.ts
Mathieu-Deharbe: Great !

src/components/utils/caseUtils.ts
Mathieu-Deharbe: Merged. Everything seems to be fine.

thangqp (approved):
Code OK , Test OK

ghazwarhili (approved):

PR #493, opened 2024-08-13
allow user to delete on creation studies

AbdelHedhili:


src/components/menus/content-contextual-menu.jsx
```diff
@@ -392,8 +392,8 @@ const ContentContextualMenu = (props) => {
     }, [selectedElements, userId]);
 
     const allowsDelete = useCallback(() => {
-        return isUserAllowed() && noCreationInProgress();
-    }, [isUserAllowed, noCreationInProgress]);
+        return isUserAllowed() && selectedElements.every((el) => el.elementUuid != null);
```
Meklo, line 394: With this change not only studies in creation will be allowed to be deleted but all types of elements, is that ok ?

AbdelHedhili: other types or elements don't have the same "oncreation" status so it doesn't affect them, we can always delete them

Meklo: Alright

Meklo (approved):

PR #498, opened 2024-08-23
Adapt the contextualMenu position.

AAJELLAL:


jamal-khey (approved):
Code OK
Small padding when we click on the icon

PR #500, opened 2024-08-28
improve access to the contextual menu.

AAJELLAL:


src/components/custom-tree-item.jsx
```diff
@@ -53,6 +87,11 @@ const CustomContent = React.forwardRef(function CustomContent(props, ref) {
             <Typography onClick={handleSelectionClick} component="div" sx={styles.label}>
                 {label}
             </Typography>
+            {hover && (
+                <Box onClick={handleAddIconClick} sx={{ marginLeft: 'auto' }}>
```
sBouzols: remove this `marginLeft: 'auto'` if you need to justify the button to the right please add `flexGrow: 1` to the preview container in the flexbox (here to put in styles.label)

src/components/custom-tree-item.jsx
```diff
@@ -41,9 +72,12 @@ const CustomContent = React.forwardRef(function CustomContent(props, ref) {
                 expanded && styles.expanded,
                 selected && styles.selected,
                 focused && styles.focused,
-                disabled && styles.disabled
+                disabled && styles.disabled,
+                hover && { backgroundColor: `${theme.aggrid.highlightColor} !important`, borderRadius: '16px' } // Add hover style with  background color
```
sBouzols: Adding an aggrid css color to a TreeItem seems a bad idea

src/components/custom-tree-item.jsx
```diff
@@ -96,6 +135,16 @@ CustomContent.propTypes = {
      * The callback to call when handle Selection Click.
      */
     onSelect: PropTypes.func,
+
+    /**
+     * The callback to call when handle Context Menu Click.
+     */
+    onContextMenu: PropTypes.func,
```
sBouzols: onAddIconClick instead of onContextMenu, it's an implementation detail that this button open the same menu as the contextual one.

src/components/tree-views-container.jsx
```diff
@@ -315,7 +315,7 @@ const TreeViewsContainer = () => {
     const mergeCurrentAndUploading = useCallback(
         (current) => {
             let elementsToMerge = Object.values(uploadingElements).filter(
-                (e) => e.directory === selectedDirectoryRef.current.elementUuid && current[e.elementName] === undefined
+                (e) => e.directory === selectedDirectoryRef.current?.elementUuid && current[e.elementName] === undefined
```
sBouzols: Not for this scope I think and why this change ?

src/components/custom-tree-item.jsx
sBouzols: seems like it's already ok then why this marginLeft ?

src/components/custom-tree-item.jsx
sBouzols: Maybe instead of propagate the button click event to the parent hierarchy, we could accept to use another `DirectoryTreeContextualMenu` here ? Because we will need another anchor setting and all.

src/components/custom-tree-item.jsx
AAJELLAL: Indeed,I don't need the marginLeft.

src/components/custom-tree-item.jsx
AAJELLAL: I agree that using aggrid css is not ideal but since the hover and aggrid will always be the same color, I tought of using it instead of creating another variable with the same value. But I can change it.

src/components/custom-tree-item.jsx
AAJELLAL: I changed onContextMenu. I don't think it's a good idea to use another DirectoryTreeContextualMenu ( a lot of code to be duplicated)

src/components/tree-views-container.jsx
AAJELLAL: I can't seem to replicate the situation where the **selectedDirectoryRef.current** is null, which caused the app to crash in this particular scenario.
However, since **selectedDirectoryRef.current** is instantiated with the **selectedDirectory**, which may be null, I believe we should maintain this check.

src/components/custom-tree-item.jsx
```diff
@@ -41,9 +72,12 @@ const CustomContent = React.forwardRef(function CustomContent(props, ref) {
                 expanded && styles.expanded,
                 selected && styles.selected,
                 focused && styles.focused,
-                disabled && styles.disabled
+                disabled && styles.disabled,
+                hover && { backgroundColor: `${theme.aggrid.highlightColor} !important`, borderRadius: '16px' } // Add hover style with  background color
             )}
             onMouseDown={handleMouseDown}
+            onMouseEnter={() => toggleHover(true)}
```
sBouzols: toogle doesn't take any argument. It's mean to "toggle" then it reverse le current boolean value.
Here if you whant to set the boolean value, then it's a set that's all.

EstherDarkish (approved):

src/components/directory-content.tsx
```diff
@@ -694,6 +695,10 @@ const DirectoryContent = () => {
                             : undefined
                     }
                     restrictMenuItems={true}
+                    anchorOrigin={{
+                        vertical: 'top',
+                        horizontal: 'right',
+                    }}
```
sBouzols: not used when anchorReference = anchorPosition
to remove

src/components/directory-tree-view.tsx
```diff
@@ -59,6 +59,13 @@ const styles = {
         fontWeight: 'inherit',
         color: 'inherit',
     }),
+    treeItemHovered: (theme: Theme) => ({
+        backgroundColor: theme.aggrid.highlightColor + '!important',
+        borderRadius: 16,
```
sBouzols:
```suggestion
        borderRadius: theme.spacing(2),
```

src/components/custom-tree-item.tsx
```diff
@@ -52,9 +82,20 @@ const CustomContent = React.forwardRef(function CustomContent(props: TreeItemCus
...
             )}
             onMouseDown={handleMouseDown}
+            /* It's not a good idea to base the hover state over those mouse events
+               because those events could be skip with html optimization.
+               However we choose to use this simple way because the
+               contextual menu opened by the AddIcon Button is shared for all the app.
+               Using the hover state directly in the CSS + isMenuOpen to manage the
+               color of this div and show/hidden state of the button caused some problems
+               when the menu is openned from another event in the app.
+            */
```
EstherDarkish:
```suggestion
            /* It's not a good idea to rely on the hover state provided by those mouse events
               because those events could be skipped with the web-browser's html optimization.
               Therefore, we choose to use this simple way because the
               contextual menu opened by the AddIcon Button is shared for all the app.
               Using the :hover CSS pseudo class and the isMenuOpen prop to determine the style can be done but
               causes issues when the menu is opened from another event in the app.
            */
```

src/components/custom-tree-item.tsx
```diff
@@ -42,6 +60,18 @@ const CustomContent = React.forwardRef(function CustomContent(props: TreeItemCus
     const handleSelectionClick = (event: React.MouseEvent<HTMLDivElement>) => {
         onSelect(nodeId);
     };
+    const handleAddIconClick: MouseEventHandler<HTMLButtonElement> = (event) => {
+        // used to open the menu
+        onAddIconClick(event, nodeId, 'anchorEl');
+    };
+
+    // We don't get a onMouseLeave event when using or leaving the contextual menu
+    // clicking outside the concerned div, then we must update the hover state manually
```
EstherDarkish:
```suggestion
    // We don't get a onMouseLeave event when using or leaving the contextual menu by
    // clicking outside the concerned div, so we must update the hover state manually.
```

souissimai (approved):
src/components/tree-views-container.tsx(655,21): error TS2322: Type '{ directory: ElementAttributes | null; open: boolean; openDialog: string; setOpenDialog: Dispatch<SetStateAction<string>>; onClose: (e: unknown) => void; ... 4 more ...; restrictMenuItems: false; }' is not assignable to type 'IntrinsicAttributes & DirectoryTreeContextualMenuProps'.
  Property 'anchorOrigin' does not exist on type 'IntrinsicAttributes & DirectoryTreeContextualMenuProps'.

src/components/tree-views-container.tsx
```diff
@@ -331,6 +341,7 @@ const TreeViewsContainer = () => {
     const mergeCurrentAndUploading = useCallback(
         (current: ElementAttributes[]): ElementAttributes[] | undefined => {
             let uploadingElementsInSelectedDirectory = Object.values(uploadingElementsRef.current).filter(
+                //TOCHECK   (e) => e.directory === selectedDirectoryRef.current?.elementUuid && current[e.elementName] === undefined // WTF ?
```
souissimai: delete this

sBouzols (approved):

PR #501, opened 2024-08-28
Fix: update selected directory's name after renaming

sBouzols:
Avoid every effects elsewhere procs when this object change instead of just its `elementUuid`.


src/components/tree-views-container.jsx
```diff
@@ -474,18 +482,6 @@ const TreeViewsContainer = () => {
                 ) {
                     dispatch(setSelectedDirectory(null));
                 }
-                if (
```
ayolab, line 477: we still have the bug when renaming root directories, I think that we need to keep this code

ayolab (changes requested):

ayolab (approved):
tests : OK
code : OK

PR #505, opened 2024-09-10
improve aggrid table

souissimai:


jamal-khey (approved):
Looks good to me.
no warn in the console

PR #516, opened 2024-09-13
prevent hiding columns when moving

souissimai:


AAJELLAL (approved):
Code and tests are ok.

PR #517, opened 2024-09-13
Add Directory move

YenguiSeddik:


src/components/menus/directory-tree-contextual-menu.tsx
```diff
@@ -377,6 +410,8 @@ const DirectoryTreeContextualMenu: React.FC<DirectoryTreeContextualMenuProps> =
                 );
             case DialogsId.ADD_NEW_CASE:
                 return <CreateCaseDialog open={true} onClose={handleCloseDialog} />;
+            case DialogsId.MOVE:
+                return <MoveDialog open={true} onClose={handleMoveDirectory} itemsCount={1} />;
```
sBouzols: Ok to use this component, but you will need to customize the `title` I think, and you need to add an option to move a directory at the root level, as describe in the ticket.

sBouzols: `onClose` should accept something like this as well I think `(node: TreeViewFinderNodeProps | null) => void;`, maybe adapt type definition in commons-ui, something like that : 
```
export interface TreeViewFinderProps {
  ...
  onClose: (nodes: TreeViewFinderNodeProps[] | TreeViewFinderNodeProps | null) => void;
  ...
```
null meaning move à root level for our case ?

src/components/tree-views-container.jsx
```diff
@@ -114,7 +114,6 @@ function updatedTree(prevRoots, prevMap, nodeId, children) {
                 };
             }
         });
-
```
sBouzols: remove this line change

src/components/tree-views-container.jsx
```diff
@@ -130,6 +129,7 @@ function updatedTree(prevRoots, prevMap, nodeId, children) {
         ...Array.prototype.concat(
             ...prevUuids
                 .filter((u) => !nextUuids.has(u))
+                .filter((uuid) => !(nodeId && prevMap[uuid] && prevMap[uuid].parentUuid !== nodeId)) // if the element is moved but not deleted
```
sBouzols: I don't think we should change this really sensible code indeed.
What's the problem here ?

src/components/menus/directory-tree-contextual-menu.tsx
souissimai: need to grey out the name of the folder being moved  preventing its selection while allowing the user to choose a new folder

souissimai (commented):
maybe need to rename the title and button of move dialog and change it "move folder" instead of "move element"

souissimai:
it seems good,
small remark : need to change button name to "Move folder" instead of "Move element"

jamal-khey (approved):

src/components/menus/directory-tree-contextual-menu.tsx
```diff
@@ -385,6 +425,15 @@ const DirectoryTreeContextualMenu: React.FC<DirectoryTreeContextualMenuProps> =
                 );
             case DialogsId.ADD_NEW_CASE:
                 return <CreateCaseDialog open={true} onClose={handleCloseDialog} />;
+            case DialogsId.MOVE_DIRECTORY:
+                return (
+                    <MoveDialog
+                        open={true}
+                        onClose={handleMoveDirectory}
+                        itemsCount={1}
+                        titleId={'moveDirectoryTitle'}
```
sBouzols: we must add the name of the directory to move in the title I think! it's not specified in the feature request but you can move a not selected folder then it's better to show the user which folder will be moved. 
* Use `activeDirectory` to get the name
* Add a translation with a variable
* give a `title` prop in MoveDialog which could be a `FormattedMessage` component not only a key, see intl documentation to add `values` to set variable in a message

YenguiSeddik: done

src/components/dialogs/move-dialog.tsx
```diff
@@ -14,10 +14,17 @@ interface MoveDialogProps {
...
+    directoryName?: string;
 }
 
-const MoveDialog: FunctionComponent<MoveDialogProps> = ({ open, onClose, itemsCount, titleId }) => {
+const MoveDialog: FunctionComponent<MoveDialogProps> = ({
+    open,
+    onClose,
+    itemsCount,
+    isDirectoryMoving,
+    directoryName,
```
sBouzols: If you pass `validationButtonText` and `title` props here, you don't add `*directory*` props here to a generic component `MoveDialog` which are relative to only one use case (move directory instead of elements).

YenguiSeddik: done

sBouzols (approved):
Code Review OK
Tests OK
Console Warning Check OK

PR #520, opened 2024-09-17
Keep order of aggrid columns when moving 

souissimai:


src/components/directory-content-table.tsx
```diff
@@ -64,7 +68,49 @@ export const DirectoryContentTable = ({
...
+    const reorderColumns = (colDef: ColDef[], newFieldOrder: string[] | undefined): ColDef[] => {
+        const fieldIndexMap = new Map(newFieldOrder?.map((field, index) => [field, index]));
+        return colDef
+            .filter((col) => fieldIndexMap.has(col.field || ''))
+            .sort((a, b) => {
+                const indexA = fieldIndexMap.get(a.field || '') ?? -1;
+                const indexB = fieldIndexMap.get(b.field || '') ?? -1;
+                return indexA - indexB;
+            });
+    };
```
jamal-khey: this function 'renderColumns' can be defined outside the component so it's will be created only once for the entire application. because the function does not need any state or props

jamal-khey (commented):
Test Ok.
small optimisation to add

jamal-khey (approved):
Code Ok
Test Ok

PR #529, opened 2024-10-01
Grid explore columns width

souissimai:
Make columns autosizable with minWidth
adjust size of column elementName


jamal-khey (approved):
The code implementation looks good overall. However, there's an important consideration regarding the 'name' column:

**Note on 'name' column behavior:**
The 'name' column is currently ignoring the `flex` attribute due to its 'pinned' status. Only `width` and `minWidth` properties are having an effect on this column.

**Action Item:**
- [x] Confirm with the PO if this behavior for the 'name' column is acceptable or if adjustments are needed.

src/components/utils/directory-content-utils.ts
```diff
@@ -142,7 +143,7 @@ export const getColumnsDefinition = (childrenMetadata: Record<UUID, ElementAttri
         }),
         field: 'lastModifiedBy',
         cellRenderer: UserCellRenderer,
-        flex: 1,
+        flex: 1.2,
```
sBouzols: maybe keep 1.2 here and change owner column to have flex 1.2 as well to have same width for `owner` and `lastModifiedBy` which show th same Avatar component

src/components/utils/directory-content-utils.ts
```diff
@@ -74,14 +74,15 @@ export const getColumnsDefinition = (childrenMetadata: Record<UUID, ElementAttri
...
+        minWidth: 400,
     },
     {
         headerName: intl.formatMessage({
             id: 'description',
         }),
         field: 'description',
         cellRenderer: DescriptionCellRenderer,
-        flex: 1.1,
+        flex: 1.5,
```
sBouzols: maybe keep 1.2 here to have same width than `owner` and `lastModifiedBy`. An edit button is equivalent to one Avatar Chip I think

src/components/utils/directory-content-utils.ts
```diff
@@ -74,14 +74,15 @@ export const getColumnsDefinition = (childrenMetadata: Record<UUID, ElementAttri
         headerCheckboxSelection: true,
         checkboxSelection: true,
         flex: 5,
```
sBouzols: then remove flex if it's incompatible with pinned column ?
Or comment it with a good comment to not reintroduce it and explain the `minWidth` as well because it's a value in pixels and that's not so cool

sBouzols (approved):
Code Review OK
Tests Done
Some exploration using `sizeColumnsToFit` to flex `'elementName'` column first and avoid raw minWidth 400 pixel value, but AGGrid seems limited for auto resizing pinned column
Console warning check OK

PR #534, opened 2024-10-07
fix delete elements when there is a lot

EtienneLt:
fix delete elements when there are a lot


src/utils/rest-api.ts
```diff
@@ -220,12 +220,19 @@ export function deleteElement(elementUuid: string) {
...
     console.info('Deleting elements : %s', elementUuids);
-    const idsParams = getRequestParamFromList('ids', elementUuids).toString();
-    return backendFetch(PREFIX_EXPLORE_SERVER_QUERIES + `/v1/explore/elements/` + activeDirectory + '?' + idsParams, {
-        method: 'delete',
-    });
+    const chunkSize = 50;
+    for (let i = 0; i < elementUuids.length; i += chunkSize) {
+        const partitionIds = elementUuids.slice(i, i + chunkSize);
+        const idsParams = getRequestParamFromList('ids', partitionIds).toString();
+        await backendFetch(
```
sBouzols, line 232: why do you wait here, why not sending delete requests in parallel ?

src/utils/rest-api.ts
```diff
@@ -220,12 +220,19 @@ export function deleteElement(elementUuid: string) {
     });
 }
 
-export function deleteElements(elementUuids: string[], activeDirectory: string) {
+export async function deleteElements(elementUuids: string[], activeDirectory: string) {
```
sBouzols, line 223: Use Promise form for homogeneity ?

src/utils/rest-api.ts
EtienneLt: The idea is usually there will be only one request and sometimes there will be more.

EtienneLt: and to prevent to send too many requests at the same time

Mathieu-Deharbe: I understand the point but I think this is a very important decision (how the big request are cut into parts, the size of the parts, if they are simultaneous or not, if something is done between each request). I Guess it will be applied in many places ?

Do the techlead all agree on this ?

@jonenst

src/utils/rest-api.ts
```diff
@@ -220,12 +220,19 @@ export function deleteElement(elementUuid: string) {
...
     console.info('Deleting elements : %s', elementUuids);
-    const idsParams = getRequestParamFromList('ids', elementUuids).toString();
-    return backendFetch(PREFIX_EXPLORE_SERVER_QUERIES + `/v1/explore/elements/` + activeDirectory + '?' + idsParams, {
-        method: 'delete',
-    });
+    const chunkSize = 50;
```
Mathieu-Deharbe, line 225: I think this should be extracted as a constant, especially if this behaviour is supposed to be applied elsewhere (cf https://github.com/gridsuite/commons-ui/pull/598)

src/utils/rest-api.ts
sBouzols, line 232: Why we don't need to put
```
// eslint-disable-next-line no-await-in-loop
```
here like in the other PR ?

src/utils/rest-api.ts
```diff
@@ -220,12 +220,19 @@ export function deleteElement(elementUuid: string) {
...
+    for (let i = 0; i < elementUuids.length; i += chunkSize) {
+        const partitionIds = elementUuids.slice(i, i + chunkSize);
+        const idsParams = getRequestParamFromList('ids', partitionIds).toString();
```
sBouzols, line 228: We don't need to filter falsy elements here ? like in the other PR ? 
```
partitionIds.filter((id) => id) // filter falsy elements
```

src/utils/rest-api.ts
jonenst: I think it's good to split requests in chunks when everything is independant. I don't think we need to overthink it, the chunk size and the parallelism is dependent on the context. Here sequential and chunksize 50 are the good values I think. we may also add a try/catch in the future so that errors in one batch don't affect the subsequent requests

sBouzols (approved):

Mathieu-Deharbe (approved):

PR #535, opened 2024-10-08
Add cases alert when reaching threshold

Meklo:


FranckLecuyer (commented):
Tests : 
When creating a new case or new study and the user quota is exceeded, the snackbar displays the message (OK), but the case or the study creation is shown in progress, and when we change to another directory and return to the original directory, the elements of this diretory are not shown ...

Meklo:
> Tests : When creating a new case or new study and the user quota is exceeded, the snackbar displays the message (OK), but the case or the study creation is shown in progress, and when we change to another directory and return to the original directory, the elements of this diretory are not shown ...

I just noticed that behaviour too and it looks like it's not a regression, we may open a ticket about it or I can quickly troubleshoot it

FranckLecuyer (approved):

PR #536, opened 2024-10-08
fix: refactor getLink method

TheMaskedTurtle:


src/components/directory-content.tsx
```diff
@@ -325,23 +330,17 @@ const DirectoryContent = () => {
     );
 
     const getLink = useCallback(
```
anistouri: You can rename it to getStudyLink

anistouri (approved):

PR #545, opened 2024-10-18
Refactor EmptyDirectory and anchorReference strategies management

sBouzols:
refactor(EmptyDirectory): rename `openDialog` callback prop to `onCreateElementButtonClick`, it didn't open a dialog but a popover and by the way it's not the concern of EmptyDirectory to know what the parent do when we click on the createElement button

refactor(menus): Manage both anchorReference strategies for contextual menus, code factorization and types extension from mui types when easy


src/components/empty-directory.tsx
```diff
@@ -61,7 +65,7 @@ const EmptyDirectory: React.FC<EmptyFolderProps> = ({ openDialog, theme }) => {
                 <h3>
                     <FormattedMessage id={'emptyDirContent'} />
                 </h3>
-                <Button variant="contained" sx={styles.button} onClick={openDialog} endIcon={<AddIcon />}>
+                <Button variant="contained" sx={styles.button} onClick={onCreateElement} endIcon={<AddIcon />}>
```
klesaulnier: rename onCreateElement to onButtonClick / onClick ? 
we're not really creating element by clicking here

src/components/directory-content.tsx
```diff
@@ -256,6 +248,27 @@ const DirectoryContent = () => {
         event.stopPropagation();
     };
 
+    const onContextMenu = useCallback(
+        (event: any, anchorStates: AnchorStatesType = defaultAnchorStates) => {
```
klesaulnier, line 243: it's a bit unfortunate to still have "any" here, but all the contextual menus are typed this way here
to fix later

src/components/directory-content.tsx
```diff
@@ -256,6 +248,27 @@ const DirectoryContent = () => {
...
+            if (anchorStates.anchorReference === 'anchorPosition') {
+                anchorStates.anchorPosition = {
+                    top: event.clientY + constants.VERTICAL_SHIFT,
+                    left: event.clientX + constants.HORIZONTAL_SHIFT,
+                };
+            } // else anchorEl
```
klesaulnier: it's a bit difficult to understand here what case it covers
can we add a comment to explain the different cases it covers, like "click on empty dir new element / click on toolbox / right click on dir / ..."

src/components/directory-content.tsx
```diff
@@ -256,6 +248,27 @@ const DirectoryContent = () => {
...
+            setDirectoryMenuAnchorStates(anchorStates);
+            //We check if the context menu was triggered from a row to prevent displaying both the directory and the content context menus
+            const isRow = !!event.target.closest(`.${CUSTOM_ROW_CLASS}`);
+            if (!isRow) {
+                dispatch(setActiveDirectory(selectedDirectory?.elementUuid));
+                handleOpenDirectoryMenu(event);
+            } else {
+                handleOpenContentMenu(event.event);
+            }
```
klesaulnier: homogenize the code a little bit to prevent using "event" for on of the element, and "event.event" for the other

src/components/empty-directory.tsx
sBouzols: Well it's the label and the feature of the button into `EmptyDirectory` component.
Ok to use `onButtonClick` prop but we should add a prop to define the button label too to be coherent...
Does this EmptyDirectory component should be this much generic... I don't think so

It's ok for me like this, for example, if a component has a 'clear' button, like MUI DateField, then it offers a `onClear` prop, not a `onButtonClick` which we guess is the clear button (https://mui.com/x/api/date-pickers/date-field/#date-field-prop-onClear)

src/components/directory-content.tsx
sBouzols: comments added

src/components/directory-content.tsx
sBouzols: fixed

src/components/directory-content.tsx
sBouzols: Well I added the event type to `onCellContextMenu` it's a beginning...

src/components/empty-directory.tsx
klesaulnier: when I see a prop named "onCreateElement", it means it's a callback triggered when an element is created
here, it's a callback to actually create an element if I'm not wrong ?
this is why I find it confusing, it's not because it should match the "onClick" from "<Button>"

klesaulnier (approved):
Code: OK
Test: OK

src/components/empty-directory.tsx
sBouzols: deal with `onCreateElementButtonClick`

PR #558, opened 2024-10-31
Fix renamed element selection

sBouzols:


src/components/directory-content.tsx
```diff
@@ -606,6 +594,21 @@ export default function DirectoryContent() {
...
+        setIsMissingDataAfterDirChange(true);
+        setCheckedRows([]);
+    }, [selectedDirectory?.elementUuid]);
+
+    useEffect(() => {
+        setIsMissingDataAfterDirChange(false);
+        // update checkecRows ElementAttributes objects if metadata changed
+        // ex: when the user renames a selected element
+        updateCheckedRows();
+    }, [childrenMetadata, updateCheckedRows]); // this will change after switching selectedDirectory
```
jamal-khey, line 610: minor comment, i think there is no need to move the use effect in the buttom of the component.
it better to keep them right after the state and props declarations. because it's a common practice , and most of the devs expect them there.

jamal-khey (approved):

src/components/directory-content.tsx
sBouzols: There is a rule which force to put `useEffect` after dependencies callbacks definitions, this was the reason I add to move it (adding `updateCheckedRows` as a new dependency)
A component should not have a lot of `useEffect` otherwise we should cut it into multiple components.
+ Effects are considered between renders **in definition order** then it's important to see the succession of all the effects without checking all the file. That's another reason why I group them all.

I think it's a matter of code style, and we don't share any in the project for now, I personally follow the following pattern when coding a React component : 
* States declaration
* Functions and Calbacks definitions
* Effects
* JSX

souissimai (approved):

PR #567, opened 2024-11-14
Fix errors when creating folders and root folders 

sBouzols:
revert previous change (naming) but was also property key change


jamal-khey (approved):
Looks good to me.
Test Ok

SlimaneAmar (approved):

PR #577, opened 2024-12-06
Fix TreeViewsContainer : useless expand when currentPath changes

sBouzols:
Do not dispatch currentPath each time mapData change otherwise it  proc a useEffect in DirectoryTreeView which expand


EstherDarkish (approved):
Code OK Test OK

PR #582, opened 2024-12-16
Optim, add user identities cache 

sBouzols:
Add a local cache and avoid multiple fetches if identity is already known.
If some user identities fetch return errors for some sub, it will retry during navigation if an element is owned or lastly modified by this user in the directory.

### DIFF
--- a/src/utils/rest-api.ts
+++ b/src/utils/rest-api.ts
@@ -185,7 +185,11 @@ export function fetchVersion() {
 }
 
 export function fetchUsersIdentities(elementUuids: string[]) {
-    console.info('fetching users identities for elements : %s', elementUuids);
+    if (elementUuids.length === 0) {
+        return Promise.resolve();
+    }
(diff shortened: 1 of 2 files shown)


jonenst (approved):

jonenst:
code ok, untested

src/utils/rest-api.ts
```diff
@@ -185,7 +185,11 @@ export function fetchVersion() {
 }
 
 export function fetchUsersIdentities(elementUuids: string[]) {
-    console.info('fetching users identities for elements : %s', elementUuids);
+    if (elementUuids.length === 0) {
+        return Promise.resolve();
+    }
+    console.info('fetching users identities for %s elements.', elementUuids.length);
+    console.debug('fetching users identities for elements: %s', elementUuids);
```
EstherDarkish: is this debug line necessary ?

sBouzols: removed

EstherDarkish (approved):

PR #581, opened 2024-12-16
Fix PR #571

Tristan-WorkGH:
Because of `@ts-expect-ignore` _inside_ JSX tags, some components parameters that changed weren't reported by tsc.

PR #583, opened 2024-12-16
Fix Typescript errors in DirectoryContent dialogs

Tristan-WorkGH:
Fix `@ts-expect-error` for ids and hidden error of #581.


Tristan-WorkGH:
#581 merged, so include in it

PR #587, opened 2024-12-26
Case import : compute base name from case-server instead of front com…

basseche:
needs : 
https://github.com/gridsuite/commons-ui/pull/661

### DIFF
--- a/src/components/dialogs/commons/prefilled-name-input.tsx
+++ b/src/components/dialogs/commons/prefilled-name-input.tsx
@@ -42,7 +42,12 @@ export default function PrefilledNameInput({ label, name, elementType }: Readonl
 
     useEffect(() => {
         // we replace the name only if some conditions are respected
-        if (caseFile && !modifiedByUser && !apiCallErrorMessage && !caseFileErrorMessage) {
+        if (
+            caseFile &&
+            (!modifiedByUser || elementType === ElementType.CASE) &&
+            !apiCallErrorMessage &&
+            !caseFileErrorMessage
+        ) {
             const { name: caseName } = caseFile;
             if (caseName) {
                 clearErrors(name);
@@ -69,6 +74,7 @@ export default function PrefilledNameInput({ label, name, elementType }: Readonl
         clearErrors,
         name,
         snackError,
+        elementType,
     ]);
 
     return (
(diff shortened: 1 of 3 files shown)


src/components/dialogs/create-case-dialog/create-case-dialog.tsx
```diff
@@ -126,7 +126,7 @@ export default function CreateCaseDialog({ onClose, open }: Readonly<CreateCaseD
                 </Grid>
             </Grid>
             <ErrorInput name={FieldConstants.CASE_FILE} InputField={FieldErrorAlert} />
-            <UploadNewCase />
+            <UploadNewCase isNewStudyCreation />
```
ghazwarhili: Why did you add isNewStudyCreation to the create-case-dialog? The UploadNewCase component is shared between case and study, and this variable is used to distinguish between the two.

src/components/dialogs/commons/prefilled-name-input.tsx
```diff
@@ -42,8 +44,9 @@ export default function PrefilledNameInput({ label, name, elementType }: Readonl
         // we replace the name only if some conditions are respected
         if (caseFile && !modifiedByUser && !apiCallErrorMessage && !caseFileErrorMessage) {
             const { name: caseName } = caseFile;
+            const currentCaseName = getValues(name);
 
-            if (caseName) {
+            if (caseName && caseName !== currentCaseName) {
```
ghazwarhili: caseName = basename + extension (PtFige-20241120-1520-enrichi_N1.zip )
currentCaseName = basename (PtFige-20241120-1520-enrichi_N1)

caseName !== currentCaseName is always true.

ghazwarhili (changes requested):

src/components/dialogs/commons/prefilled-name-input.tsx
basseche: yes you are right. I wanted only to check if file doesn't change (we select the same file)

src/components/dialogs/create-case-dialog/create-case-dialog.tsx
basseche: ok

basseche:
> [ suggestion to fix this PR](https://github.com/gridsuite/gridexplore-app/pull/592#issue-2759542134)

The problem with this fix that : 
- it will introduce a bug, because it doesn't take into account user modification of the name, it will always calls endPoint.
- it does not solve problem of UniqueNameInput component.
- it is a refactoring it will make PrefilledNameInput component obsolete and we should remove it.

ghazwarhili:
I didn't say this is the final solution.. it's a proposal or suggestion to properly implement the new endpoint, like a first version to be completed.

thangqp (approved):

ghazwarhili (approved):

PR #592, opened 2025-01-03
Correction for case base name

ghazwarhili:
[ suggestion to fix this PR](https://github.com/gridsuite/gridexplore-app/pull/587#issue-2759542134)


basseche:
> [ suggestion to fix this PR](https://github.com/gridsuite/gridexplore-app/pull/587#issue-2759542134)

The problem with this fix that : 
- it will introduce a bug, because it doesn't take into account user modification of the name, it will always calls endPoint.
- it does not solve problem of UniqueNameInput component.
- it is a refactoring it will make PrefilledNameInput component obsolete and we should remove it.

src/components/dialogs/create-case-dialog/create-case-dialog.tsx
```diff
@@ -57,12 +58,14 @@ export default function CreateCaseDialog({ onClose, open }: Readonly<CreateCaseD
...
     const {
         formState: { errors, isValid },
+        getValues,
     } = createCaseFormMethods;
 
     const isFormValid = isObjectEmpty(errors) && isValid;
 
     const activeDirectory = useSelector((state: AppState) => state.activeDirectory);
     const userId = useSelector((state: AppState) => state.user?.profile.sub);
+    const [modifiedByUser, setModifiedByUser] = useState(false);
```
thangqp, line 68:
```suggestion
    const [nameModifiedByUser, setNameModifiedByUser] = useState(false);
```

src/components/dialogs/commons/upload-new-case.tsx
```diff
@@ -23,12 +24,13 @@ const MAX_FILE_SIZE_IN_MO = 100;
 const MAX_FILE_SIZE_IN_BYTES = MAX_FILE_SIZE_IN_MO * 1024 * 1024;
 
 export default function UploadNewCase({
+    modifiedByUser = false,
```
thangqp, line 27:
```suggestion
    nameModifiedByUser = false,
```

src/components/dialogs/commons/upload-new-case.tsx
```diff
@@ -58,6 +60,24 @@ export default function UploadNewCase({
...
     }, [caseFileLoading, caseFileName]);
 
+    const fetchBaseName = (currentFile: any) => {
+        const { name: currentCaseFileName } = currentFile;
+        const name = isNewStudyCreation ? FieldConstants.STUDY_NAME : FieldConstants.CASE_NAME;
+        if (currentCaseFileName && !modifiedByUser) {
+            getBaseName(currentCaseFileName)
+                .then((response) => {
+                    setValue(name, response, {
+                        shouldValidate: true,
```
basseche, line 70: you should not use shouldValidate : true, it will removes check on duplicate name

basseche (changes requested):
shouldValidate: true will remove check on file name unicity and should not be used

src/components/dialogs/create-study-dialog/create-study-dialog.tsx
```diff
@@ -77,7 +77,7 @@ export default function CreateStudyDialog({ open, onClose, providedExistingCase
     const activeDirectory = useSelector((state: AppState) => state.activeDirectory);
     const selectedDirectory = useSelector((state: AppState) => state.selectedDirectory);
     const userId = useSelector((state: AppState) => state.user?.profile.sub);
-
+    const [modifiedByUser, setModifiedByUser] = useState(false);
```
basseche, line 80: it is not a good thing to put the state of the child here, you render the whole form when the state of the child changes !

PR #594, opened 2025-01-09
Description can be modified directly from contingency list edition dialog

basseche:


src/components/dialogs/contingency-list/contingency-list-utils.ts
```diff
@@ -85,11 +87,15 @@ export const getExplicitNamingFormDataFromFetchedElement = (response: any) => {
     }
     return {
         [FieldConstants.EQUIPMENT_TABLE]: result,
+        [FieldConstants.DESCRIPTION]: description,
+        [FieldConstants.NAME]: name,
     };
 };
 
-export const getScriptFormDataFromFetchedElement = (response: any) => ({
+export const getScriptFormDataFromFetchedElement = (response: any, name: string, description: string) => ({
```
Mathieu-Deharbe, line 95: Might be better to keep the same parameter order for all those similar functions (`getCriteriaBasedFormDataFromFetchedElement`, `getExplicitNamingFormDataFromFetchedElement`, `getScriptFormDataFromFetchedElement`).

src/components/dialogs/contingency-list/contingency-list-utils.ts
```diff
@@ -85,11 +87,15 @@ export const getExplicitNamingFormDataFromFetchedElement = (response: any) => {
...
     [FieldConstants.SCRIPT]: response.script,
+    [FieldConstants.NAME]: name,
+    [FieldConstants.DESCRIPTION]: description,
```
Mathieu-Deharbe, line 98: Also is there a reason to use || here : `[FieldConstants.DESCRIPTION]: description || '',` in `getCriteriaBasedFormDataFromFetchedElement` but neither in `getExplicitNamingFormDataFromFetchedElement` or `getScriptFormDataFromFetchedElement`.

src/components/dialogs/contingency-list/creation/contingency-list-creation-dialog.tsx
```diff
@@ -31,7 +32,7 @@ import { AppState } from '../../../../redux/types';
 
 const schema = yup.object().shape({
     [FieldConstants.NAME]: yup.string().trim().required('nameEmpty'),
-    [FieldConstants.DESCRIPTION]: yup.string().max(500, 'descriptionLimitError'),
+    [FieldConstants.DESCRIPTION]: yup.string().max(MAX_CHAR_DESCRIPTION, 'descriptionLimitError'),
```
Mathieu-Deharbe: If I enter a description above 500 characters I can't validate, which is fine, but the message `'descriptionLimitError'` never appears. While `nameEmpty` appears under the Name input when the name is indeed empty.

Same problem in the simple description modification and filter description modification so nothing new in this ticket. Still, that seems to be a bug.

src/components/dialogs/contingency-list/edition/script/script-edition-dialog.tsx
```diff
@@ -101,7 +106,11 @@ export default function ScriptEditionDialog({
             id: contingencyListId2,
             script: contingencyList[FieldConstants.SCRIPT],
         };
-        return saveScriptContingencyList(newScript, contingencyList[FieldConstants.NAME]);
+        return saveScriptContingencyList(
+            newScript,
+            contingencyList[FieldConstants.NAME],
+            contingencyList[FieldConstants.DESCRIPTION] ?? ''
```
Mathieu-Deharbe, line 112: Aren't it going to save a description as an empty string ? If yes wouldn't it be better to keep description "undefined" and to not send it to the back in case of undefined instead of sending an empty string ?

src/utils/rest-api.ts
```diff
@@ -573,9 +574,10 @@ export function saveCompositeModification(id: string, name: string) {
  * @returns {Promise<Response>}
  */
 export function saveCriteriaBasedContingencyList(id: string, form: CriteriaBasedEditionFormData) {
-    const { name, equipmentType, criteriaBased } = form;
+    const { name, description, equipmentType, criteriaBased } = form;
     const urlSearchParams = new URLSearchParams();
     urlSearchParams.append('name', name);
+    urlSearchParams.append('description', description ?? '');
```
Mathieu-Deharbe, line 580: Why send description in this case ? If description is undefined, simply ignore it.
```suggestion
if (description) {
        urlSearchParams.append('description', description);
    }
```
(same for the 3 following similar functions)

Mathieu-Deharbe (changes requested):
Contingency list : 
Capture d’écran du 2025-01-15 10-24-26

Naming filter : 

Capture d’écran du 2025-01-15 10-24-45

Padding should be the same. (The naming filter version looks better in my opinion) (same with the scripted contingency list)

src/utils/rest-api.ts
basseche: because when you want to delete description, you send an empty one.

src/components/dialogs/contingency-list/edition/script/script-edition-dialog.tsx
basseche: the type is string you can't give undefined to string that's why we should do this check and i think if we send undefined to back it will really save undefined word.

src/components/dialogs/contingency-list/contingency-list-utils.ts
basseche: done

src/components/dialogs/contingency-list/creation/contingency-list-creation-dialog.tsx
basseche: This bug does not seem to be new maybe new. I can try to resolve it anyway.

basseche:
> Contingency list : Capture d’écran du 2025-01-15 10-24-26
> 
> Naming filter :
> 
> Capture d’écran du 2025-01-15 10-24-45
> 
> Padding should be the same. (The naming filter version looks better in my opinion) (same with the scripted contingency list)

done

src/components/dialogs/contingency-list/creation/contingency-list-creation-dialog.tsx
basseche: it seems that this comes from ExpandingTextField in commons-ui :

`       helperText: <Typography variant="caption">{descriptionCounter}</Typography>,`
       
 I'm not sure this should be solved here. 
 It's clear that message is useless if we keep it like that.

basseche: The solution would be to remove the message

Mathieu-Deharbe (approved):

PR #607, opened 2025-01-27
Developer mode

EstherDarkish:
Uses https://github.com/gridsuite/commons-ui/pull/682


src/components/dialogs/user-settings-dialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+        marginTop: theme.spacing(1),
+        flexGrow: 1,
+    }),
+    controlItem: {
+        flexGrow: 1,
+    },
+    parameterLine: {
+        display: 'flex',
+    },
+    parameterValue: {},
```
AAJELLAL: To be deleted?

src/components/dialogs/user-settings-dialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Dialog to display user settings
+ * @param open Is the dialog open ?
+ * @param onClose Event to close the dialog
+ */
+export default function UserSettingsDialog({ open, onClose }: Readonly<UserSettingsDialogProps>) {
+    const dispatch = useDispatch();
+
+    const [enableDeveloperMode, handleChangeEnableDeveloperModeLocal] = useParameterState(PARAM_DEVELOPER_MODE);
```
AAJELLAL:
```suggestion
    const [enableDeveloperMode, handleChangeEnableDeveloperMode] = useParameterState(PARAM_DEVELOPER_MODE);
```

src/components/dialogs/user-settings-dialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const [developerMode, setDeveloperMode] = useState<boolean>(enableDeveloperMode);
+
+    const handleValidate = () => {
+        dispatch(selectEnableDeveloperMode(developerMode));
+        handleChangeEnableDeveloperModeLocal(developerMode);
+        onClose();
+    };
+
+    const handleClose = (_: SyntheticEvent, reason?: string) => {
+        if (reason === 'backdropClick') {
```
AAJELLAL: Consider defining 'backdropClick' as a constant at the top of the file or within an 'utils' file.

src/redux/types.d.ts
```diff
@@ -56,6 +56,7 @@ export type UploadingElement = {
 export interface AppState extends CommonStoreState {
     [PARAM_THEME]: GsTheme;
     [PARAM_LANGUAGE]: GsLang;
+    enableDeveloperMode: boolean;
```
souissimai, line 59: [PARAM_DEVELOPER_MODE]

src/components/dialogs/user-settings-dialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+                </Box>
+                {developerMode && (
+                    <Alert severity="warning">
+                        <FormattedMessage id="DeveloperModeWarningMsg" />
+                    </Alert>
+                )}
+            </DialogContent>
+            <DialogActions>
+                <CancelButton onClick={handleClose} />
+                <Button onClick={handleValidate} disabled={false} variant="outlined">
```
AAJELLAL: why disabled is set to false? don't we need to disable it if no changes  have occurred. Otherwise, the call is triggered when clicking 'Validate', even if no changes have been made

src/components/app-top-bar.tsx
```diff
@@ -45,6 +46,10 @@ export default function AppTopBar({ userManagerInstance }: Readonly<AppTopBarPro
 
     const [languageLocal, handleChangeLanguage] = useParameterState(PARAM_LANGUAGE);
 
+    const [enableDeveloperModeLocal] = useParameterState(PARAM_DEVELOPER_MODE);
```
AAJELLAL:
```suggestion
    const [enableDeveloperMode] = useParameterState(PARAM_DEVELOPER_MODE);
```

src/components/dialogs/user-settings-dialog.tsx
souissimai: parameterValue ? to delete

src/components/app.tsx
```diff
@@ -70,6 +70,9 @@ export default function App() {
                         // TODO remove cast when prototype is fixed in commons-ui
                         dispatch(selectComputedLanguage(getComputedLanguage(param.value) as GsLangUser));
                         break;
+                    case PARAM_DEVELOPER_MODE:
+                        dispatch(selectEnableDeveloperMode(String(param.value) === 'true'));
```
souissimai, line 74: In my opinion : keep value string in this case as argument no need to define it boolean and then convert it to string

src/components/dialogs/user-settings-dialog.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <Dialog fullWidth open={open} onClose={handleClose}>
+            <DialogTitle>
+                <FormattedMessage id="UserSettings" />
+            </DialogTitle>
+            <DialogContent>
+                <Box sx={styles.parameterLine}>
+                    <Box sx={styles.parameterName}>
+                        <FormattedMessage id="EnableDeveloperMode" />
+                    </Box>
+                    <Box sx={styles.parameterValue}>
```
souissimai: no style

AAJELLAL (commented):
Some fixes to do according to the design:
- Add more spacing between the parameter and the <Alert severity="warning"> .
 - The font families used are not identical to the requested design."
 
- Open 2 tabs, initiate the developer mode change popup in both. 
  After changing the mode in tab 1, tab 2 doesn't reflect the change until the popup is closed and reopened

src/components/dialogs/user-settings-dialog.tsx
```diff
@@ -0,0 +1,94 @@
+/**
```
souissimai: this dialog is the same for explore and study , move it to commons ui

src/components/app.tsx
EstherDarkish: I would like to have the same way to handle the variable as in gridstudy, to simplify and unify things between the two applications.

src/components/dialogs/user-settings-dialog.tsx
EstherDarkish: Thanks !

src/redux/types.d.ts
EstherDarkish: I had a lot of trouble with Typescript with this, so I'll keep as is for now.

src/components/dialogs/user-settings-dialog.tsx
EstherDarkish: Thanks

src/components/dialogs/user-settings-dialog.tsx
EstherDarkish: Yes, thanks

src/components/dialogs/user-settings-dialog.tsx
EstherDarkish: I won't do it to keep the code consistent with all the other instances of this backdropClick test.

src/components/dialogs/user-settings-dialog.tsx
EstherDarkish: Thanks

EstherDarkish:
> Some fixes to do according to the design:
> 
>     * Add more spacing between the parameter and the  .
> 
>     * The font families used are not identical to the requested design."
> 
>     * Open 2 tabs, initiate the developer mode change popup in both.
>       After changing the mode in tab 1, tab 2 doesn't reflect the change until the popup is closed and reopened

src/components/app-top-bar.tsx
EstherDarkish: Done

EstherDarkish:
> Some fixes to do according to the design:
> 
>     * Add more spacing between the parameter and the  .
Github seems to have removed something from your comment. Anyway, this dialog is temporary as the long term vision of the UX Designer is to put the developer mode next to the language and theme selectors, when we would have implemented the developer roles in GridAdmin.
> 
>     * The font families used are not identical to the requested design."
The requested design is a quick mockup. We do not want serif fonts anywhere.
> 
>     * Open 2 tabs, initiate the developer mode change popup in both.
>       After changing the mode in tab 1, tab 2 doesn't reflect the change until the popup is closed and reopened
Thanks, it is fixed.

src/components/dialogs/user-settings-dialog.tsx
EstherDarkish: I'd rather not as the logic needed uses the reducer and I do not want to put that in the topbar.

src/components/dialogs/user-settings-dialog.tsx
EstherDarkish: This dialog is temporary and do not need the same level of validations as the others. If we do not want to submit the value, we can use the cancel button.

src/components/dialogs/user-settings-dialog.tsx
AAJELLAL: I agree with Maissa. Given the duplication of the code, it’s better to create a component in commons-ui.
 I’m not entirely sure I understood your response. You just need to create a component with handleValidate as a prop, so the component remains independent of the business logic. 
Using handleValidate, you can pass whatever you want.

AAJELLAL (commented):
The synchronization issue between two tabs is still present in Gridexplore and GridStudy.

AAJELLAL (commented):
Code and test are ok.

AAJELLAL (approved):
code and test are ok.

src/utils/config-params.ts
```diff
@@ -10,8 +10,9 @@ export const APP_NAME = 'Explore';
 
 export const PARAM_THEME = 'theme';
 export const PARAM_LANGUAGE = 'language';
+export const PARAM_DEVELOPER_MODE = 'enableDeveloperMode';
 
-const COMMON_CONFIG_PARAMS_NAMES = new Set([PARAM_THEME, PARAM_LANGUAGE]);
+const COMMON_CONFIG_PARAMS_NAMES = new Set([PARAM_THEME, PARAM_LANGUAGE, PARAM_DEVELOPER_MODE]);
```
sBouzols: change to remove then. Will be moved to common parameters later

sBouzols (approved):
Code Review OK
Tests OK
Console warning check OK

PR #651, opened 2025-04-07
Rework deferred fetches hooks and error handling

Meklo:
- Rework `useDeferredFetch` and `useMultipleDeferredFetch` to fix and simplify them
  - Getting rid of their internal reducers
  - Mobilize `Promise.allSettled` for `useMultipleDeferredFetch` to synchronize fetches 
  - Replace parsing functions which retrieve error message ids in function of error code by predefined maps
- Centralize in `rest-errors.ts` error handling functions which are common to `ContentToolbar`, `ContentContextualMenu` and `DirectoryTreeContextualMenu`
- Decompose multi-step error checking functions into multiple conditions to enhance readability


src/components/utils/rest-errors.ts
```diff
@@ -4,56 +4,160 @@
...
-            messageId: 'deleteConflictError',
-        });
+        snackError({ messageId: 'deleteConflictError' });
         return true;
     }
     return false;
 };
+
+export const handlePasteError = (error: CustomError, intl: IntlShape, snackError: SnackError) => {
+    const message = generatePasteErrorMessages(intl)[4156];
```
FranckLecuyer: 4156 ?

FranckLecuyer (changes requested):
Tests: When moving an element in a directory we can't write into, the snackbar shows a generic message not very explicit : 'Problem encountered while moving an element. Not moved : eec34372-1ea4-4534-b957-c35fcb9edac1' ...
Instead, we should see this message : "You cannot move this element to the targeted folder. Unauthorized action"

src/components/utils/rest-errors.ts
Meklo: hmmm I have no idea how this got here, replaced it by `error.status` as expected

FranckLecuyer (approved):

PR #657, opened 2025-04-14
listen announcement ws

AbdelHedhili:


src/components/app-top-bar.tsx
Tristan-WorkGH, line 1: I don't think the alert is "in" the topbar ...

Meklo (approved):

PR #661, opened 2025-04-17
change export network url from case server to conversion server

EtienneLt:
change the url to convert a case to another format as the code was migrate from powsybl network case server to powsybl network conversion server


antoinebhs (approved):
Code OK

basseche (approved):

PR #684, opened 2025-06-19
Add datatest-id for directories actions

etiennehomer:


AmirHdm (approved):
All implemented data-testid attributes adhere to the conventional naming convention.

PR #742, opened 2025-10-13
Upgrade ViteJS to version 7

FranckLecuyer:


Meklo (approved):
All OK you can just bump vite to newly released 7.1.11 version
No need to upgrade commons ui i think

PR #756, opened 2025-11-04
Display directory owner identity in the directory properties dialog

FranckLecuyer:
## PR Summary

In the directory properties dialog, instead of displaying the directory owner, as it is stored in the directory-server database, we display now the directory owner identity, to be consistent with what is displayed in the tooltip for a directory element.

### DIFF
--- a/src/components/dialogs/directory-properties/directory-properties-dialog.tsx
+++ b/src/components/dialogs/directory-properties/directory-properties-dialog.tsx
@@ -25,6 +25,7 @@ import {
     hasManagePermission,
     PermissionDTO,
     PermissionType,
+    fetchUsersIdentities,
 } from '../../../utils/rest-api';
 import {
     Group,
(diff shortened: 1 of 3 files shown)


src/components/dialogs/directory-properties/directory-properties-dialog.tsx
```diff
@@ -119,11 +121,32 @@ function DirectoryPropertiesDialog({ open, onClose, directory }: Readonly<Direct
...
+                        `${res?.data?.[directory.owner]?.firstName} ${res?.data?.[directory.owner]?.lastName}`
+                    );
+                }
+            })
+            .catch((error) => {
+                snackError({
+                    messageTxt: error.message,
+                    headerId: 'directoryOwnerIdentityFetchError',
+                });
+                setDirectoryOwnerIdentity(undefined);
```
Meklo: Not needed I think ? As directoryOwnerIdentity has `directory?.owner` for default value there's no need to unset it if the fetch fails I guess

FranckLecuyer: :+1: 
Removed

Meklo (approved):

PR #768, opened 2025-11-20
Check write permission before study opening

dbraquart:
## PR Summary
Currently a GS user can open any element, including a study, in a directory where the write permission is restricted. This user can open a study and perform any operations, like node deletions for instance.
To protect the studies in read-only directories, a check will be made: no openning if no write permission.

### DIFF
--- a/src/components/directory-content-dialog.tsx
+++ b/src/components/directory-content-dialog.tsx
@@ -73,7 +73,7 @@ function DirectoryContentDialog(
 ) {
     const intl = useIntl();
     const dispatch = useDispatch();
-    const { snackError, snackWarning } = useSnackMessage();
+    const { snackError } = useSnackMessage();
     const itemSelectionForCopy = useSelector((state: AppState) => state.itemSelectionForCopy);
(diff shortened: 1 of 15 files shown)


src/components/directory-content-dialog.tsx
```diff
@@ -176,7 +178,9 @@ function DirectoryContentDialog(
                         case ElementType.STUDY: {
                             const url = getStudyUrl(event.data.elementUuid);
                             if (url) {
-                                window.open(url, '_blank');
+                                if (selectedDirectoryWritable) {
+                                    window.open(url, '_blank');
+                                }
```
Meklo, line 183: How about showing a snack info to indicate the element is not accessible in addition with the change of pointer so that it's extra clear to the user why their cursor is crossed ?

dbraquart: finally, after discusssions with PO+UX, there is no feedback (pop-up, snack msg, tooltip) for now. Just the not-allowed pointer.

Meklo (approved):

PR #775, opened 2025-12-01
enhance byFilter contingency list

EtienneLt:

### DIFF
--- a/src/components/dialogs/contingency-list/filter-based/contingency-list-filter-based-form.tsx
+++ b/src/components/dialogs/contingency-list/filter-based/contingency-list-filter-based-form.tsx
@@ -11,6 +11,7 @@ import {
     ElementType,
     EquipmentType,
     FieldConstants,
+    GridSection,
     MuiStyles,
     OverflowableChipWithHelperText,
     ResizeHandle,
@@ -18,7 +19,6 @@ import {
 } from '@gridsuite/commons-ui';
 import { Grid, useMediaQuery, useTheme } from '@mui/material';
 import { useSelector } from 'react-redux';
-import { FormattedMessage } from 'react-intl';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
 import { ImperativePanelGroupHandle, Panel, PanelGroup } from 'react-resizable-panels';
@@ -149,9 +149,7 @@ export default function ContingencyListFilterBasedForm({
                         <Grid item>
                             <DescriptionField />
                         </Grid>
-                        <Grid item component="h4">
-                            <FormattedMessage id="Filters" />
-                        </Grid>
+                        <GridSection title="Filters" />
                         <Grid item xs>
                             <DirectoryItemsInput
                                 titleId="FiltersListsSelection"
@@ -179,6 +177,7 @@ export default function ContingencyListFilterBasedForm({
                 <FilterBasedContingencyListVisualizationPanel
                     isDataOutdated={isDataOutdated}
                     setIsDataOutdated={setIsDataOutdated}
+                    filters={filters}
                 />
             </Panel>
         </PanelGroup>
--- a/src/components/dialogs/contingency-list/filter-based/contingency-list-filter-based-form.tsx
+++ b/src/components/dialogs/contingency-list/filter-based/contingency-list-filter-based-form.tsx
@@ -11,7 +11,6 @@ import {
     ElementType,
     EquipmentType,
     FieldConstants,
-    GridSection,
     MuiStyles,
     OverflowableChipWithHelperText,
     ResizeHandle,
@@ -22,6 +21,7 @@ import { useSelector } from 'react-redux';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
 import { ImperativePanelGroupHandle, Panel, PanelGroup } from 'react-resizable-panels';
+import { FormattedMessage } from 'react-intl';
 import { AppState } from '../../../../redux/types';
 import { ContingencyFieldConstants, FilterElement, FilterSubEquipments } from '../../../../utils/contingency-list.type';
 import { FilterBasedContingencyListVisualizationPanel } from './filter-based-contingency-list-visualization-panel';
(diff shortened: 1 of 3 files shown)


src/components/dialogs/contingency-list/filter-based/contingency-list-filter-based-form.tsx
```diff
@@ -149,9 +149,7 @@ export default function ContingencyListFilterBasedForm({
                         <Grid item>
                             <DescriptionField />
                         </Grid>
-                        <Grid item component="h4">
-                            <FormattedMessage id="Filters" />
-                        </Grid>
```
EtienneLt, line 154: if it is better I can revert and set h3

src/components/dialogs/contingency-list/filter-based/filter-based-contingency-list-visualization-panel.tsx
```diff
@@ -178,27 +180,50 @@ export function FilterBasedContingencyListVisualizationPanel(
 
     const studyName = selectedFolder ? selectedFolder + separator + selectedStudy : selectedStudy;
 
+    const formatPathName = useCallback((name: string) => {
+        console.log(name.length);
```
flomillot: I think you forgot this log

src/components/dialogs/contingency-list/filter-based/contingency-list-filter-based-form.tsx
flomillot: GridSection is not even an `item` so IMO yes
Or maybe add the option in the component ?
But here we didn't use GridItem and GridSection at all, and now I think there is too much space between the title and the field, but maybe ask the PO to have his opinion

src/components/dialogs/contingency-list/filter-based/filter-based-contingency-list-visualization-panel.tsx
```diff
@@ -178,27 +180,50 @@ export function FilterBasedContingencyListVisualizationPanel(
 
     const studyName = selectedFolder ? selectedFolder + separator + selectedStudy : selectedStudy;
 
+    const formatPathName = useCallback((name: string) => {
```
flomillot: Maybe store directly the formated name with a use memo instead of using a callback?

src/components/dialogs/contingency-list/filter-based/filter-based-contingency-list-visualization-panel.tsx
```diff
@@ -178,27 +180,50 @@ export function FilterBasedContingencyListVisualizationPanel(
...
+
     return (
         <Grid container direction="column" sx={{ height: '100%' }}>
-            <Grid item component="h3" padding={0.5}>
-                <FormattedMessage id="visualization" />
-            </Grid>
+            <GridSection title="visualization" />
             <Grid item container alignItems="center" justifyContent="space-between">
-                <Grid item>
+                <Grid item xs={1} paddingTop={1}>
```
flomillot: It's just my opinion but I prefer how it was before, we were able to see more.
Maybe we can find an inbetween ? Check with PO to be sure ?

Image

Image

src/components/dialogs/contingency-list/filter-based/filter-based-contingency-list-visualization-panel.tsx
```diff
@@ -55,12 +56,13 @@ const getRowId = (params: GetRowIdParams<RowData>) => {
 export type FilterBasedContingencyListVisualizationPanelProps = {
     isDataOutdated: boolean;
     setIsDataOutdated: (value: boolean) => void;
+    filters: FilterElement[];
```
flomillot: By passing this prop you re render these components every time `filters` change or you just need to know if there is one or more filters, so maybe create variable `hasFilters`  in the parent component and pass it instead of the whole filters

src/components/dialogs/contingency-list/filter-based/filter-based-contingency-list-visualization-panel.tsx
```diff
@@ -178,27 +180,50 @@ export function FilterBasedContingencyListVisualizationPanel(
...
                     )}
                 </Grid>
                 <Grid item>
-                    <Button onClick={() => setIsOpen(true)} variant="contained" color="primary" component="label">
-                        <FormattedMessage id="selectStudyDialogButton" />
+                    <Button
+                        disabled={filters === undefined || filters.length === 0}
+                        onClick={() => setIsOpen(true)}
+                        variant={selectedStudy.length > 0 ? 'contained' : undefined}
+                        color="primary"
```
flomillot: You can remove this line it's already the color

flomillot (commented):
IMO too much empty space between the title the icon the study name and the button

Image

src/components/dialogs/contingency-list/filter-based/filter-based-contingency-list-visualization-panel.tsx
```diff
@@ -180,43 +179,51 @@ export function FilterBasedContingencyListVisualizationPanel(
...
 
-    const formatPathName = useCallback((name: string) => {
-        console.log(name.length);
-        if (name.length > 40) {
-            const splitNameList = name.split('/');
+    const formatPathName = useMemo(() => {
+        if (studyName.length > 48) {
+            const splitNameList = studyName.split('/');
+            const lastFolder = splitNameList.at(-1);
+            if (lastFolder === undefined) {
```
flomillot: How is it possible to enter in this condition?

From my test :

Image

src/components/dialogs/contingency-list/filter-based/filter-based-contingency-list-visualization-panel.tsx
```diff
@@ -180,43 +179,51 @@ export function FilterBasedContingencyListVisualizationPanel(
...
+                return studyName;
+            }
             if (splitNameList.length > 2) {
-                if (splitNameList[0].length + splitNameList[splitNameList.length - 1].length < 40) {
-                    return `${splitNameList[0]}/.../${splitNameList[splitNameList.length - 1]}`;
+                const firstFolder = splitNameList.at(0);
```
flomillot: Same I don't understand how this can be undefined

flomillot (approved):

PR #783, opened 2025-12-11
block edit equipment type for filter outside dev mode

EtienneLt:
## PR Summary
editing a filter and change its equipment type can provoc many problems.
now it will be allow only in dev mod


flomillot (approved):

flomillot:
Merged into https://github.com/gridsuite/gridexplore-app/pull/786

PR #811, opened 2026-02-04
[GRD-3838] make upload file size limit configurable via apps-metadata

khouadrired:
This PR moves the upload file size limit from a hardcoded value to the apps-metadata configuration.

The maximum allowed file size is now dynamically retrieved from the Explore app metadata (maxFileSizeInMb).
If the metadata is missing, a default value of 100 MB is used to ensure backward compatibility.

### DIFF
--- a/src/components/dialogs/commons/upload-new-case.tsx
+++ b/src/components/dialogs/commons/upload-new-case.tsx
@@ -7,26 +7,33 @@
 
 import { ChangeEvent, useCallback, useMemo, useState } from 'react';
 import { FormattedMessage, useIntl } from 'react-intl';
+import { useSelector } from 'react-redux';
 import { Button, CircularProgress, Grid, Input } from '@mui/material';
 import { useController, useFormContext } from 'react-hook-form';
 import { ErrorInput, FieldConstants, FieldErrorAlert } from '@gridsuite/commons-ui';
 import type { UUID } from 'node:crypto';
 import { HTTP_CONNECTION_FAILED_MESSAGE, HTTP_UNPROCESSABLE_ENTITY_STATUS } from 'utils/UIconstants';
 import { createCaseWithoutDirectoryElementCreation, deleteCase } from '../../../utils/rest-api';
+import type { AppState, ExploreMetadata } from '../../../redux/types';
 
 export interface UploadNewCaseProps {
     isNewStudyCreation?: boolean;
     getCurrentCaseImportParams?: (uuid: UUID) => void;
 }
 
-const MAX_FILE_SIZE_IN_MO = 100;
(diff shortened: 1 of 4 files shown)


src/redux/types.d.ts
```diff
@@ -78,3 +78,8 @@ export interface AppState extends CommonStoreState {
     itemSelectionForCopy: ItemSelectionForCopy;
     reorderedColumns: string[];
 }
+
+export type ExploreMetadata = Metadata & {
+    name: 'Explore';
+    MaxFileSizeInMo?: number;
+};
```
thangqp:
```suggestion

export type ExploreMetadata = Metadata & {
    name: 'Explore';
    maxFileSizeInMb?: number;
};
```
We should move this type declaration to commons-ui:
https://github.com/gridsuite/commons-ui/blob/main/src/utils/types/metadata.ts#L22

In fact, this type describes the apps-metadata.json artifact used at deployment time, which does not depend on gridexplore-app.

khouadrired: okey will do that.

src/components/dialogs/commons/upload-new-case.tsx
```diff
@@ -7,26 +7,35 @@
...
     isNewStudyCreation = false,
     getCurrentCaseImportParams,
 }: Readonly<UploadNewCaseProps>) {
     const intl = useIntl();
+    const MAX_FILE_SIZE_IN_MB = useSelector((state: AppState) => {
+        const exploreApp = state.appsAndUrls.find((metadata) => metadata.name === 'Explore') as ExploreMetadata;
+        return exploreApp?.maxFileSizeInMb ?? DEFAULT_MAX_FILE_SIZE_IN_MB;
+    });
+
+    const MAX_FILE_SIZE_IN_BYTES = MAX_FILE_SIZE_IN_MB * 1024 * 1024;
```
thangqp:
```suggestion
    const appsAndUrls = useSelector((state: AppState) => state.appsAndUrls);
    
    const maxFileSizeInMb = useMemo(() => {
        const exploreMetadata = appsAndUrls.find(isExploreMetadata);
        return (exploreMetadata?.maxFileSizeInMb ?? DEFAULT_MAX_FILE_SIZE_IN_MB);
    }, [appsAndUrls])
    
    const maxFileSizeInByte = maxFileSizeInMb * 1024 * 1024;
```
I suggest to create a predicate in commons-ui, i.e. isExploreMetadata, and use it to compute max size.

I see yours solution works without any performance problem in this case, in which appsAndUrls never changes. However, computing a value (even never changed) may lead to a miss understanding for someone who is not familiar to using useSelector. He/she can use your pattern to compute a complex value such as an object/array.

khouadrired: i see your point , will apply that.

thangqp (approved):

PR #827, opened 2026-03-09
chore: use load modification form from commons UI

dbraquart:
## PR Summary
Second part of https://github.com/gridsuite/gridexplore-app/pull/824 to support modification case.


achour94 (changes requested):
The Labels are not the same, all the fields are shown as optionnal !

achour94 (approved):

PR #841, opened 2026-03-25
feat: add optional removal flag to modification dialogs

achour94:

### DIFF
--- a/src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
+++ b/src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
@@ -68,7 +68,14 @@ const styles = {
 
 type SpecificModificationDialogProps = Pick<
     ModificationDialogProps<any, any>,
-    'formSchema' | 'dtoToForm' | 'formToDto' | 'errorHeaderId' | 'titleId' | 'ModificationForm' | 'isModification'
+    | 'formSchema'
+    | 'dtoToForm'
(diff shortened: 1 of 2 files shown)


src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
```diff
@@ -68,7 +68,14 @@ const styles = {
...
     ModificationDialogProps<any, any>,
-    'formSchema' | 'dtoToForm' | 'formToDto' | 'errorHeaderId' | 'titleId' | 'ModificationForm' | 'isModification'
+    | 'formSchema'
+    | 'dtoToForm'
+    | 'formToDto'
+    | 'errorHeaderId'
+    | 'titleId'
+    | 'ModificationForm'
+    | 'isModification'
+    | 'removeOptional'
```
dbraquart, line 78: because removeOptional is related to the modification case, there is no need to add a new prop: we can use existing prop "isModification", and set it to true for all modification dialogs ?

dbraquart (commented):
code : a suggestion made
tests: ok

dbraquart (approved):

src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
achour94: i prefer keeping them as separate explicit props

PR #845, opened 2026-03-31
Allow opening and editing the byFilter modification form in GridExplore

benrejebmoh:
## PR Summary
Allow opening and editing the byFilter modification form in GridExplore


dbraquart (commented):
code:ok
Tests:
- "Edited field" values are not translated
- EmptyField / ValueOrEmptyField not translated
- I guess some "RTE" translations will be needed. ex: Tension d'enroulement 1 (kV) vs Tension d'enroulement Or (kV)
A change in gridexplore-app-rte (gitlab) may be required in src/plugins/translations/fr_rte.json
- The dialog width is very smaller than in grid-study. Maybe we have to ask TL/PO if it is acceptable ?

dbraquart (approved):

PR #850, opened 2026-04-15
Get element names from directory-server

carojeandat:


src/utils/contingency-list.type.ts
```diff
@@ -33,13 +33,21 @@ export interface FilterBasedContingencyList {
         equipmentTypes: string[];
     }>;
 }
+export interface FilterBasedContingencyListBackend {
```
thangqp: FilterBasedContingencyListBackend is unused

thangqp: to remove

thangqp (commented):
? fetchElementNames(filterIds).then((elementNames) => {
TO
? fetchElementNames(filterIds).then((elementNames: Record<string, string>) => {

const filterIds = new Set(contingencyList.filters?.map((filter: FilterAttributes) => filter.id)) as Set<string>;
TO
const filterIds = new Set(contingencyList.filters?.map((filter: FilterAttributes) => filter.id));

I can not comment directly in the code so I write my usggestion here

thangqp (commented):
using filterIds > 0 to narrowing two subtypes of contingency list is not a clean way to check whether a filter based or id based contingency list.
The right way is based on fileld `type`  as in getContingencyUriParamType(contingencyListType)

src/utils/contingency-list.type.ts
carojeandat: done

thangqp (approved):

thangqp:
> using filterIds > 0 to narrowing two subtypes of contingency list is not a clean way to check whether a filter based or id based contingency list. The right way is based on fileld `type` as in getContingencyUriParamType(contingencyListType)

Correct typing will be done in the next PR: https://github.com/gridsuite/gridexplore-app/pull/862

antoinebhs (approved):
lgtm

PR #862, opened 2026-05-11
Rework on typing contingency list at api with enrichement 

thangqp:


src/utils/rest-api.ts
```diff
@@ -543,27 +545,52 @@ export function createFilterBasedContingency(
...
-              };
-          })
-        : Promise.resolve(contingencyList);
-}
-
-export function getContingencyList(type: string, id: string) {
+export function getContingencyList<T extends keyof ContingencyListTypeMap>(
+    type: T,
+    id: string
+): Promise<ContingencyListTypeMap[T]>;
```
thangqp:
```suggestion
export function getContingencyList(type: 'FILTERS', id: string): Promise<FilterBasedContingencyList>;
export function getContingencyList(type: 'IDENTIFIERS', id: string): Promise<PrepareContingencyListForBackend>;
```

thangqp: It is better to expicitly typing without an abstract layer with type mapper if only some types in the union

carojeandat (approved):

carojeandat (approved):

src/utils/rest-api.ts
```diff
@@ -535,27 +536,57 @@ export function createFilterBasedContingency(
 }
 
 /**
- * Get contingency list by type and id
- * @returns {Promise<Response>}
+ * Enhances a given contingency list with additional data such as element names if applicable.
```
antoinebhs: too much commentS??

thangqp: DONE! Rectified

PR #868, opened 2026-05-18
Make name and description umodifiable in composite modification form

basseche:
## PR Summary

Editing name and description from composite dialog take long time, since the whole modification is saved, the user can think that his modification didn't work and try to edit again before the request is done.
We disable this way of editing, user can always edit name and description from outside the composite dialog.

### DIFF
--- a/src/components/dialogs/network-modification/composite-modification/composite-modification-form.tsx
+++ b/src/components/dialogs/network-modification/composite-modification/composite-modification-form.tsx
@@ -7,8 +7,11 @@
 
 import { FieldConstants, TextInput, unscrollableDialogStyles } from '@gridsuite/commons-ui';
 import { Box } from '@mui/material';
+import { useFormContext } from 'react-hook-form';
 
 export default function CompositeModificationForm() {
+    const { getValues } = useFormContext();
+    const name = getValues(FieldConstants.NAME);
     return (
         <>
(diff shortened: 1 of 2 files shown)


src/components/dialogs/network-modification/composite-modification/composite-modification-form.tsx
```diff
@@ -5,31 +5,32 @@
...
+                <TextInput
                     name={FieldConstants.NAME}
-                    label="nameProperty"
-                    elementType={ElementType.MODIFICATION}
-                    activeDirectory={activeDirectory}
+                    label="name"
+                    formProps={{
+                        disabled: true,
+                        sx: {
+                            width: '50ch',
```
Meklo: With names too long it now cuts off, maybe we should keep the previous width or add an tooltip + ellipse to still be able to display the full name

basseche: It was a demand from the designer, i will ask him for the tooltip.

basseche: Fixed, minWidth set to 50 characters and otherwise we take string size limited to 100%

Meklo: All good

Meklo (approved):

PR #869, opened 2026-05-19
navigate on directory breadcrumbs click / directory deletion

AbdelHedhili:
## PR Summary
the url was not updating when navigating through the directories breadcrumbs


flomillot (changes requested):
There is many other places where we dispatch the selected directory, but we don't change the URL. I think we should include it in the dispatch to be sure it's done every time.

AbdelHedhili:
> There is many other places where we dispatch the selected directory, but we don't change the URL. I think we should include it in the dispatch to be sure it's done every time.

I added the navigate when we change the selectedDirectory (so it fixed the delete directory not updating the url also) but I didn't put it inside the dispatch. We never did things like that before we keep the dispatch only for redux data update so i'd rather leave it that way

flomillot (approved):

PR #870, opened 2026-05-20
Update network-modification table API

flomillot:
## Summary

Adapts `CompositeModificationDialog` to the network-modification table that now lives in commons-ui, and enables inline renaming of composite modifications from that table.

- Migrates to the current `NetworkModificationsTable` API: the table takes a pre-built `columns` array instead of the former `createAllColumns` factory; the name column uses commons-ui's `NameHeaderRenderer` / `NameCellRenderer`.
- The name column's `onChange` callback renames a composite modification through `updateNetworkModificationsMetadata` (added in the commons-ui companion PR): a modification nested inside a composite is a network modification, not a directory element, so its metadata is updated directly on the network-modification-server.
- Adds the `networkModificationRenamingError` message shown by the cell on failure.

Companion PRs:
- commons-ui: https://github.com/gridsuite/commons-ui/pull/1132
- gridstudy-app: https://github.com/gridsuite/gridstudy-app/pull/3945

### DIFF
--- a/src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
+++ b/src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
@@ -4,7 +4,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
-import { SetStateAction, useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import Box from '@mui/material/Box';
 import { useForm } from 'react-hook-form';
@@ -25,7 +25,6 @@ import {
     loadModificationFormToDto,
     LoadForm,
     ModificationType,
-    NameHeaderProps,
     NetworkModificationMetadata,
     NO_ITEM_SELECTION_FOR_COPY,
     snackWithFallback,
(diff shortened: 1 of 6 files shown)


src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
```diff
@@ -115,20 +115,19 @@ interface CompositeModificationDialogProps {
...
-const createBaseColumns = (
-    isRowDragDisabled: boolean,
-    modificationsCount: number,
-    nameHeaderProps: NameHeaderProps,
-    _setModifications: React.Dispatch<SetStateAction<ComposedModificationMetadata[]>>
-): ColumnDef<ComposedModificationMetadata>[] => [
+// Rename a (nested) composite modification. It is a network modification inside the parent composite
+// (not a directory element), so its metadata is updated directly on the network-modification-server.
+const renameCompositeModification = (modification: ComposedModificationMetadata, newName: string) =>
+    updateNetworkModificationsMetadata([modification.uuid], { name: newName, type: modification.type });
```
Meklo, line 121: I noticed if the nested composite is initially disabled and we rename it, it switches to being enabled which I'm not sure we want

flomillot: ah yes, good catch
Actually we have a guard in the back end. 
https://github.com/gridsuite/network-modification-server/blob/73809a380f207bb9e6f65bc89c4641db280a0861/src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java#L721-L723
But it's totally useless because in the DTO we have 
https://github.com/gridsuite/network-modification/blob/239d61e5eb799f6c054081820f3e81870471c18f/src/main/java/org/gridsuite/modification/dto/ModificationInfos.java#L118-L120
I am gonna fix it in network modification

Meklo (approved):

PR #888, opened 2026-06-05
chore(GeneratorModification): use form from commons UI

dbraquart:


KoloMenek (approved):
Tested locally, works perfectly !

PR #904, opened 2026-06-24
Scroll to routed root directory at gridexplore opening

flomillot:
Deep-linking to a root directory (or any already-loaded directory) goes through the **fast path** of the URL resolution effect (`tree-views-container.tsx`), which only selected the directory without scrolling the tree to it. As a result, opening GridExplore on a routed root element did not scroll the tree to it.

This drives the tree scroll declaratively: the URL resolution effect sets a `directoryToScroll` state from **both** the fast path and the slow path, and a dedicated effect performs the delayed `scrollIntoView`, so routed root elements are scrolled into view too.

The scroll effect's cleanup cancels the pending timeout when the target changes or on unmount, so an older pending scroll can no longer fire after a newer directory selection.

Also increase the delay to 700.

### DIFF
--- a/src/components/tree-views-container.tsx
+++ b/src/components/tree-views-container.tsx
@@ -594,7 +594,6 @@ export default function TreeViewsContainer({ sourceItemUuid }: { readonly source
 
     const { loadPath } = useDirectoryPathLoader();
 
-    // The directory the tree should scroll to. Setting it (re)schedules the scroll declaratively below.
     const [directoryToScroll, setDirectoryToScroll] = useState<UUID | undefined>(undefined);
 
     useEffect(() => {
@@ -603,8 +602,6 @@ export default function TreeViewsContainer({ sourceItemUuid }: { readonly source
         }
         // Even if the directoryHtmlElement below exists, there are still new renders that will break the scroll on nested directories.
         // Using a delay with timeout is not clean but is the only short and working solution we found.
-        // The cleanup cancels the pending scroll when the target changes or on unmount, so only the
-        // latest directory selection can trigger scrollIntoView.
         const timeout = setTimeout(() => {
             document.getElementById(directoryToScroll)?.scrollIntoView({
                 behavior: 'smooth',
--- a/src/components/tree-views-container.tsx
+++ b/src/components/tree-views-container.tsx
@@ -594,17 +594,26 @@ export default function TreeViewsContainer({ sourceItemUuid }: { readonly source
 
     const { loadPath } = useDirectoryPathLoader();
 
-    const scrollTreeToDirectory = useCallback((directoryUuid: UUID) => {
+    // The directory the tree should scroll to. Setting it (re)schedules the scroll declaratively below.
+    const [directoryToScroll, setDirectoryToScroll] = useState<UUID | undefined>(undefined);
+
+    useEffect(() => {
+        if (!directoryToScroll) {
+            return undefined;
+        }
         // Even if the directoryHtmlElement below exists, there are still new renders that will break the scroll on nested directories.
         // Using a delay with timeout is not clean but is the only short and working solution we found.
-        setTimeout(() => {
-            document.getElementById(directoryUuid)?.scrollIntoView({
+        // The cleanup cancels the pending scroll when the target changes or on unmount, so only the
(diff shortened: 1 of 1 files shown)


src/components/tree-views-container.tsx
```diff
@@ -594,6 +594,18 @@ export default function TreeViewsContainer({ sourceItemUuid }: { readonly source
...
+        // Even if the directoryHtmlElement below exists, there are still new renders that will break the scroll on nested directories.
+        // Using a delay with timeout is not clean but is the only short and working solution we found.
+        setTimeout(() => {
+            document.getElementById(directoryUuid)?.scrollIntoView({
+                behavior: 'smooth',
+                block: 'center',
+                inline: 'nearest',
+            });
+        }, 700);
+    }, []);
```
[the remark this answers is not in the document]
ghazwarhili: The setTimeout has no cleanup, I confirm this remark

src/components/tree-views-container.tsx
```diff
@@ -594,6 +594,18 @@ export default function TreeViewsContainer({ sourceItemUuid }: { readonly source
...
+    const scrollTreeToDirectory = useCallback((directoryUuid: UUID) => {
+        // Even if the directoryHtmlElement below exists, there are still new renders that will break the scroll on nested directories.
+        // Using a delay with timeout is not clean but is the only short and working solution we found.
+        setTimeout(() => {
+            document.getElementById(directoryUuid)?.scrollIntoView({
+                behavior: 'smooth',
+                block: 'center',
+                inline: 'nearest',
+            });
+        }, 700);
```
ghazwarhili, line 615: Extract 700 into constant SCROLL_TREE_DELAY_MS for example

src/components/tree-views-container.tsx
```diff
@@ -594,6 +594,18 @@ export default function TreeViewsContainer({ sourceItemUuid }: { readonly source
 
     const { loadPath } = useDirectoryPathLoader();
 
+    const scrollTreeToDirectory = useCallback((directoryUuid: UUID) => {
+        // Even if the directoryHtmlElement below exists, there are still new renders that will break the scroll on nested directories.
+        // Using a delay with timeout is not clean but is the only short and working solution we found.
+        setTimeout(() => {
```
ghazwarhili: Raising the timeout value from 500 to 700 is still a fragile workaround in my view, but I don't have a better alternative for now.

src/components/tree-views-container.tsx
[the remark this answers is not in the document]
flomillot: I changed the way to scroll

src/components/tree-views-container.tsx
flomillot: Why creating a constant for something used only once ?

src/components/tree-views-container.tsx
flomillot: Yes I know, working on it

src/components/tree-views-container.tsx
ghazwarhili: using a named constant makes the code cleaner and more readable at the call site than using a bare numeric value

flomillot: I dont agree with that, for me its boilerplate code. Without constant you see on one sight the exact value.

ghazwarhili: Anyway, this is something we commonly do in our code to keep the code clean
For the same reason, for example:
const DELAY_BEFORE_WEBSOCKET_CONNECTED = 12000;
export const DEFAULT_TIMEOUT_MS = 50_000;
and there are plenty of other similar examples

flomillot: image

flomillot: And DEFAULT_TIMEOUT_MS is used several times. For the first one for me i'ts not a good practice and dont make the code cleaner.
@etiennehomer what do you think, should always create a const when we use settimeout ?

ghazwarhili (approved):

src/components/tree-views-container.tsx
etiennehomer: We can follow sonar rules for such questions. So not required to use a constant if only one use

PR #907, opened 2026-06-26
Migrate to Grid2

antoinebhs:

### DIFF
--- a/src/components/dialogs/commons/upload-new-case.tsx
+++ b/src/components/dialogs/commons/upload-new-case.tsx
@@ -8,7 +8,7 @@
 import { ChangeEvent, useCallback, useMemo, useState } from 'react';
 import { FormattedMessage, useIntl } from 'react-intl';
 import { useSelector } from 'react-redux';
-import { Button, CircularProgress, Grid, Input } from '@mui/material';
+import { Button, CircularProgress, Grid2 as Grid, Input } from '@mui/material';
 import { useController, useFormContext } from 'react-hook-form';
 import {
     ErrorInput,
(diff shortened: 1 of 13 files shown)


src/components/dialogs/commons/upload-new-case.tsx
```diff
@@ -140,7 +140,7 @@ export default function UploadNewCase({ getCurrentCaseImportParams }: Readonly<U
     return (
         <>
             <Grid container alignItems="center" spacing={1} pt={1}>
```
klesaulnier, line 142: In previous PR, I remember we had to set width 100% / size 12 in order to keep old behaviour of "container" grids, which was to be full sized

Not tested here though

src/components/dialogs/delete-dialog.tsx
```diff
@@ -104,21 +95,13 @@ export default function DeleteDialog({
...
-                        id={gridSimpleDeleteFormatMessageId}
-                        values={{ itemName: <span>{gridItems.length === 1 && renderElement(gridItems)}</span> }}
-                    />
-                </Grid>
-            </Grid>
+            <FormattedMessage
+                tagName="span"
+                id={gridSimpleDeleteFormatMessageId}
+                values={{ itemName: <span>{gridItems.length === 1 && renderElement(gridItems)}</span> }}
+            />
```
klesaulnier, line 104: Didn't test, but has the layout remained the same ?

klesaulnier (approved):
Code ok, if possible just check comments before merging in order to reduce potential test errors

src/components/dialogs/delete-dialog.tsx
antoinebhs: The layout remained the same, I've tried where possible to remove unnecessary Grid usages

src/components/dialogs/commons/upload-new-case.tsx
antoinebhs: We decided to limit setting width 100% / size 12 only where it was necessary. Often when it's included in a container/grid of the correct size or child items are already with a set size, it's not necessary. Checking the layout for this one showed that it's not necessary.

PR #909, opened 2026-06-30
Move the generate-template button above the explicit naming table

flomillot:
## Change

Follow-up to gridsuite/commons-ui#1219 (CSV buttons rework). On the explicit naming contingency list form:

- Render the contained **"Generate template"** button on its own row above the table (it no longer comes from the bottom toolbar), next to the import button.
- The imported file name truncates (single line + tooltip) so it no longer pushes the row to a new line.
- Restore the row spacing lost when the root `Grid` container was replaced by a `Stack`: a nested `Grid` container only inherits the spacing CSS variables from a parent `Grid` container, so the row now declares `spacing` explicitly.

## Dependency

Requires gridsuite/commons-ui#1219 (exported `CsvDownloadButton`, `CsvProps` without `getTemplateData`) to be merged and released first.

### DIFF
--- a/src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.tsx
+++ b/src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.tsx
@@ -134,7 +134,7 @@ export default function ExplicitNamingForm() {
             <Grid>
                 <DescriptionField />
             </Grid>
-            <Grid container justifyContent="space-between" alignItems="center">
+            <Grid container spacing={2} justifyContent="space-between" alignItems="center">
                 <Grid>
                     <CsvDownloadButton
                         data={getTemplateData}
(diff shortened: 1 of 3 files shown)


src/components/dialogs/contingency-list/explicit-naming/explicit-naming-form.tsx
```diff
@@ -130,11 +131,20 @@ export default function ExplicitNamingForm() {
...
                 />
             </Grid>
+            <Grid>
+                <DescriptionField />
+            </Grid>
             <Grid container justifyContent="space-between" alignItems="center">
                 <Grid>
-                    <DescriptionField />
+                    <CsvDownloadButton
+                        data={getTemplateData}
```
ghazwarhili, line 140: it should be getTableData here ?

flomillot: Why ?

basseche (approved):

ghazwarhili (approved):

PR #915, opened 2026-07-07
chore(LineModification): use form from commons-UI

dbraquart:
## PR Summary

LineModification form is now supported in composite modifications dialog. Requires https://github.com/gridsuite/commons-ui/pull/1232


achour94 (approved):
code and test ok

PR #918, opened 2026-07-16
Display elements deletion in progres

Meklo:
## PR Summary

Needs https://github.com/gridsuite/commons-ui/pull/1242

### DIFF
--- a/src/components/utils/renderers/name-cell-renderer.tsx
+++ b/src/components/utils/renderers/name-cell-renderer.tsx
@@ -17,20 +17,31 @@ import {
     OverflowableText,
 } from '@gridsuite/commons-ui';
 
-const waitingForAsyncCreation = (metadata: ElementAttributes, objectType: ElementType) =>
+const DELETING_STATUS = 'DELETING';
+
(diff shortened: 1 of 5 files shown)


src/components/utils/renderers/name-cell-renderer.tsx
```diff
@@ -17,20 +17,31 @@ import {
     OverflowableText,
 } from '@gridsuite/commons-ui';
 
-const waitingForAsyncCreation = (metadata: ElementAttributes, objectType: ElementType) =>
+const DELETING_STATUS = 'DELETING';
```
flomillot: import from commons ui

flomillot:
Also there is no notification on the folder's name. For me we should see the same notification and block the different actions (same for elements).
It should also be impossible to select it in the table.

Meklo:
Oups I've just seen your comment about interactions/notifications now, I'm right on it

That said I agree we should block actions but making it unselectable would introduce an asymetry between how creating and deleting elements are handled

flomillot (approved):

PR #926, opened 2026-07-22
Edit a loadflow process config

carojeandat:


src/components/directory-content-dialog.tsx
```diff
@@ -509,15 +510,29 @@ function DirectoryContentDialog(
...
                         processConfigId={currentProcessConfigId}
                         open
                         onClose={handleCloseProcessConfigDialog}
                         name={elementName}
                         description={activeElement.description}
                         directory={activeDirectory}
-                        fetchSAProcessConfig={fetchProcessConfig}
-                        updateSAProcessConfig={updateSAProcessConfig}
+                        fetchProcessConfig={fetchProcessConfig}
+                        updateProcessConfig={updateProcessConfig}
```
klesaulnier, line 535: to simplify gridexplore code we could use a single entrypoint for all processconfig
if we add a third process config type, we would not even have to modify this code if we do so

klesaulnier (approved):
Test: OK

PR #929, opened 2026-07-24
Set default compression to GZIP for XIIDM, BIIDM, JIIDM formats

FranckLecuyer:

### DIFF
--- a/src/components/dialogs/export-case-dialog.tsx
+++ b/src/components/dialogs/export-case-dialog.tsx
@@ -55,6 +55,10 @@ enum CompressionType {
 const compressions = [CompressionType.ZIP, CompressionType.GZIP];
 const onlyZipCompression = [CompressionType.ZIP];
 const CGMES_FORMAT = 'CGMES';
+const XIIDM_FORMAT = 'XIIDM';
+const BIIDM_FORMAT = 'BIIDM';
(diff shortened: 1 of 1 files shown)


src/components/dialogs/export-case-dialog.tsx
```diff
@@ -196,6 +200,8 @@ export default function ExportCaseDialog({ selectedElements, onClose, onExport }
                                     selectedCompression !== CompressionType.ZIP
                                 ) {
                                     setSelectedCompression(CompressionType.ZIP);
+                                } else if (defaultGzipFormats.includes(event.target.value)) {
+                                    setSelectedCompression(CompressionType.GZIP);
```
thangqp, line 199:
```suggestion
                                if (defaultGzipFormats.includes(event.target.value)) {
                                    setSelectedCompression(CompressionType.GZIP);
                                } else {
                                    setSelectedCompression(CompressionType.ZIP);
                                }
```
Same as in study: https://github.com/gridsuite/gridstudy-app/pull/4110/changes#r3645187924
In addition, we could remove the check current seleted compression since it's a react state (if reset same value, no additional notif)

FranckLecuyer: :+1:

thangqp (approved):

PR #954, opened 2026-08-13
Editing VoltageLevelSectionCreationForm from in gridexplore

benrejebmoh:


src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
```diff
@@ -414,6 +418,18 @@ export default function CompositeModificationDialog({
                         removeOptional: false,
                     },
                 ],
+                [
+                    ModificationType.CREATE_VOLTAGE_LEVEL_SECTION,
+                    {
+                        formSchema: voltageLevelSectionCreationFormSchema,
+                        dtoToForm: voltageLevelSectionCreationDtoToForm,
+                        formToDto: (form) => voltageLevelSectionCreationFormToDto(form),
```
KoloMenek, line 483: Doesn't need the (form) => voltageLevelSectionCreationFormToDto(form), it could be just voltageLevelSectionCreationFormToDto

KoloMenek (changes requested):
Tested locally and it works as expected. Small comment but not sure if it is a good practice or not so it can be ignored.

src/components/dialogs/network-modification/composite-modification/composite-modification-dialog.tsx
benrejebmoh: Not necessary, i just followed another form

KoloMenek (approved):

PR #965, opened 2026-08-31
use universal usetabs hook

khouadrired:


thangqp (approved):
Test OK in the scope of refactorisation.

However, we need to create a ticket in order to use useTabs for all hypothesis which are multi-tab, creattion type and modifcation type

PR #968, opened 2026-09-04
add sharing status

souissimai:

PR #975, opened 2026-09-15
Keep directory content column widths on table refresh

flomillot:
## PR Summary

**Before:** any action in a directory (rename, move, create, notification…) reset the column widths set by the user in the elements table.

The column definitions are rebuilt on every render of `DirectoryContent` and passed as a new `columnDefs` array, so AG Grid re-applies `width` and `flex` from them. This became visible on the *Name* column with #968: it switched from `flex: 2` (a no-op on a pinned column, which kept the user width) to `width: 300`, re-applied on every refresh.

**After:** columns use `initialWidth` / `initialFlex`, only applied when the columns are created. Widths set by the user are kept across refreshes, and columns that were not resized still flex with the grid width.


ghazwarhili (approved):
code => OK
test => OK

